PR #2, opened 2019-11-22
My version

djbj:
Link to my deployed project:
https://vigorous-bhabha-cb1b0d.netlify.com/
Just getting started again so just did the bare minimum. Had trouble hiding the results of the survey when the survey onLoads for the first time. Did not style in detail. Maybe I should have made a surveyComponent and mount it to app.js instead of putting all the logic inside app.js

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,133 @@
-import React from 'react'
+import React, { useState } from 'react'
+
+const ageGroups = ["<20", "21-40", "40+"]
 
 export const App = () => {
+  const [personName, setPersonName] = useState("");
+  const [skiing, setSkiing] = useState(false);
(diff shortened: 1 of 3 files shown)


code/src/App.js
```diff
@@ -1,9 +1,135 @@
...
+      <div id="survey">
+
+      <form
+        className="question-form"
+         //onLoad={event => document.getElementById('review').style.visibility = 'hidden'}
+        onSubmit={(event) => {
+          event.preventDefault();
+          document.getElementById('survey').style.display = 'none';
+          document.getElementById('review').style.visibility = 'visible'}
+        }
```
JennieDalgren: Try to stick to "React solutions" if possible.
For example add a state that takes care of submitted true or false and that renders components or not depending on that.

PR #3, opened 2019-11-29
Christmas Survey

johannawaern:
Link to my deployed project:
https://christmas-survey.netlify.com/

Really fun project. As always, difficult to understand where to start bc I don't know React :)But great tutorials from Jennie and great code along with Damian on Wednesday helped me out! 

Not that I can think of.


code/src/App.js
```diff
@@ -1,9 +1,148 @@
...
+            <input
+              id='name'
+              type='text'
+              onChange={event => setName(event.target.value)}
+              required
+              value={name}
+            />
+
+            <div className='divider'>
+              * * * * * * * * * * * * * * * * * * * * *
```
ericaglimsholt: Note that is someone has help reading the page (like a screen reader) it will way "*" here. So maybe it's better to add this to the CSS instead - then it will not read it

code/src/App.js
```diff
@@ -1,9 +1,148 @@
...
+              required
+              value={nickname}
+            />
+
+            <div className='divider'>
+              * * * * * * * * * * * * * * * * * * * * *
+            </div>
+
+            <h2>Which Christmas Character is your spirit animal?</h2>
+            {characters.map(group => (
```
ericaglimsholt: The map function takes an array of `characters` here and outputting a `character`, so maybe that's a better name than `group`?

code/src/App.js
```diff
@@ -1,9 +1,148 @@
...
+                {group}
+              </label>
+            ))}
+
+            <div className='divider'>
+              * * * * * * * * * * * * * * * * * * * * *
+            </div>
+
+            <h2>What do you prefer to eat?</h2>
+            {foods.map(group => (
```
```suggestion
            {foods.map(food => (
```

code/src/App.js
```diff
@@ -1,9 +1,148 @@
...
+              <option value='Driving Home For Christmas'>
+                Driving Home For Christmas - Chris Rea
+              </option>
+              <option value='I Don´t Believe in Santa Claus'>
+                I Don´t Believe in Santa Claus - the Vandals
+              </option>
+              <option value='Last Christmas'>Last Christmas - Wham</option>
+              <option value='All I Want For Christmas Is You'>
+                All I Want For Christmas Is You - Mariah Carey
+              </option>
```
ericaglimsholt: Here you also can create an array with these values and use `map()` for rendering this

PR #4, opened 2019-11-29
Theresa Ulwahn - Project 10 The Survey

TheresaUlwahn:
Link to my deployed project:
https://mrsucodesproject10thesurvey.netlify.com/

Worked together with the team as much as possible and it gave result of being able to finish earlier then previous weeks. 
My code works perfetcly and the survey hides upon the command. Though when I deployed it, it doesnt follow. Karol helped me see that the submit text code is placed in the middel of the code, which I cant fint on the VSC.. So strange and down know how to solve that. Even if so, super proud over my site! 
Skärmavbild 2019-11-29 kl  08 23 48
Skärmavbild 2019-11-29 kl  08 24 04

Things I'd like to have clarified or explained in more detail: It feels like the info about starting up the projects could be more detailed. Also to know that NOT to as last week when we heard that it was better to make many components. Upon doing that and then on Wednesday hearing that was not the purpose, it would have been good to know.

Wish you a lovely weekend and that you all get well from illness!

Kindly,
Theresa from Miami


ericaglimsholt:
Hi Theresa!
I agree, something is a bit wierd here. When I downloaded you repo I got an error saying that `background-image: linear-gradient(rgba(0, 0, 0, 0.5), rgba(0, 0, 0, 0.5)), url("./Components/Pictures/ringdips.jpg");` doesn't match the file `Ringdips.jpg`. But in Netlify it's right (with a big R) - so it seems like your repository and Netlify isn't synced for some reason... 

When I run you project locally I see that `useState` works perfect as you are describing. But we need to fix this issue with Netlify... Maybe you already have tried this (sorry if this is annoying!) but have you try logging out (`console.log`) your `setSubmit()` to see if it goes from `false` => `true`? As should be happening when you click the button. And because the issue is on Netlify so you need to upload it there and try it... Not as smooth sadly :(

PR #6, opened 2019-11-29
Project 10 - Sofie Nyblad

nyblad:
https://nyblad-quiz.netlify.com/

Felt good overall, but in the same time feels like I don't get it yet since I didn't manage to do all I wanted to. I put all code but the summary in App.js and felt like I would have needed more time to do components out of the questions, started but felt it messed up most of the conditions I had and didn't know how to fix that.

- How to do this with conditional rendering - tried but didn't figure it out, in Summary.js:

```
const refreshPage = () => {
window.location.reload()
}
<button className="restart" onClick={refreshPage}>
```

- Is there any better solution for not submitting the form when hitting enter after typing in name in first question? Tried with `disabled={question !== 4}` on the submitbutton first but didn't work so googled and found the solution I have now. 

- How to work with the aria attributes when mapping the radiobuttons? One option should have `aria-checked:true` and `tabIndex: -1` I believe..? Couldn't make the keyboard navigation work correct with my custom styling, are able to tab between radiobuttons but not move with arrows or check one.


code/src/App.js
```diff
@@ -1,9 +1,171 @@
...
+
+  //Function to handle next-button/onclick
+  const handleNext = () => {
+    setQuestion(question + 1)
+  }
+
+  //Function to handle next-button/onclick
+  const handleBack = () => {
+    setQuestion(question - 1)
+  }
```
ericaglimsholt: You can merge this into one function! So if you have one for handle the navigation, you can send `-1` or `+1` when you are clicking on the button. Then you can have a function like this:

```
const handleNavigation = (direction) => {
   setQuestion(direction)
}
```

code/src/App.js
```diff
@@ -1,9 +1,171 @@
...
+                <label>
+                  <h2># Please enter your name to participate *</h2>
+                  <input
+                    type="text"
+                    required
+                    placeholder="Type your name here..."
+                    onChange={(event) => setText(event.target.value)}
+                    value={text}
+                    // Prevent form to submit if hitting enter after typing name
+                    onKeyPress={(event) => { event.key === "Enter" && event.preventDefault() }}
```
ericaglimsholt: Nice one!

code/src/App.js
```diff
@@ -1,9 +1,171 @@
...
+
+            {question === 2 && (
+              <div className="question">
+                <label>
+                  <h2>#1: What's the best location?</h2>
+                  <div className="select-main">
+                    <select
+                      onChange={event => setSelect(event.target.value)}
+                      value={select}
+                    >
```
ericaglimsholt: Add `required` to the section?

code/src/App.js
```diff
@@ -1,9 +1,171 @@
...
+  const [question, setQuestion] = useState(1)
+  const [text, setText] = useState("")
+  const [select, setSelect] = useState("")
+  const [radioButton, setRadioButton] = useState()
+  const [checkbox, setCheckbox] = useState()
+  const [range, setRange] = useState(5)
+  const [submitted, setSubmitted] = useState(false)   //False because form is not submitted from start (initial value)
+
+  //Array with values for mapping radio buttons in #3
+  const radioButtons = ["Apple", "Banana", "Minions?"]
```
```suggestion
  const fruits = ["Apple", "Banana", "Minions?"]
```

code/src/components/Summary.js
```diff
@@ -0,0 +1,24 @@
...
+      <h2>Thank you {props.text}!</h2>
+      <p className="user-answer">#1: {props.select}</p>
+      <p className="correct-answer"> Vacayay of course! Or home sweet home <span role="img" aria-label="house">🏡</span></p>
+      <p className="user-answer">#2: {props.radioButton}</p>
+      <p className="correct-answer">Minions crave for <span role="img" aria-label="banana">🍌</span></p>
+      <p className="user-answer">#3: {props.checkbox}</p>
+      <p className="correct-answer">Yes, sloths are slow. The slowest.</p>
+      <p className="user-answer">#4: {props.range}</p>
+      <p className="correct-answer">Depending on legend Santa has 8 or 9 <span role="img" aria-label="deer">🦌</span></p>
+      <button className="restart" onClick={refreshPage}>Again?</button>
```
ericaglimsholt: This works?

code/src/App.js
nyblad: Can add that, wanted to limit the required ones while testing over and over again.

code/src/components/Summary.js
nyblad: Yeah it works, but was told that I could try something more "reacty" than refresh the page. But couldn't figure out how. I leave it as it is.

code/src/App.js
nyblad: Will fix this!

PR #8, opened 2019-11-30
Survey project - Disa Forshaw-de Beau

Disa-san:
Link to my deployed project: https://survey-form-disa.netlify.com

My reflections on how this week's project turned out: I really enjoyed working on this project and I found it to be just the right amount of challenging. As always, I could have done more work on it, but had to decide a point at which to stop and move on. 

Things I'd like to have clarified or explained in more detail: Working with more than one component.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -18,113 +18,114 @@ export const App = () => {
   }
 
   return (
-    <div className="result">
-      {showResult && (
-        <div>
-          <h2>So, I think I know you now:</h2>
-          <p> You are a {fun === "fun" ? "fun" : "no fun"} {ageGroup} {animal} and your best friend is {bestFriend}.</p>
-        </div>
+    <div className="survey-wrapper">
+      <div className="result">
+        {showResult && (
+          <div>
+            <h2>So, I think I know you now:</h2>
+            <p> You are a {fun === "fun" ? "fun" : "no fun"} {ageGroup} {animal} and your best friend is {bestFriend}.</p>
+          </div>
 
-      )}
+        )}
 
 
 
-      {!showResult && (
-        <form onSubmit={handleSubmit}>
+        {!showResult && (
+          <form onSubmit={handleSubmit}>
 
-          <section className="questions">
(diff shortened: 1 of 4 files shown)


code/src/App.js
```diff
@@ -1,9 +1,130 @@
...
+
+
+const ageGroups = ["young", "old", "perfect"]
 
 export const App = () => {
+  const [ageGroup, setAgeGroup] = useState("perfect")
+  const [animal, setAnimal] = useState("")
+  const [showResult, setShowResult] = useState(false)
+  const [bestFriend, setBestFriend] = useState("")
+  const [fun, setFun] = useState("fun")
```
ericaglimsholt: It looks like this can be a bool?
```suggestion
  const [fun, setFun] = useState(true)
```

code/src/App.js
```diff
@@ -1,9 +1,130 @@
...
+
+            <div className="spiritAnimal">
+
+              <h3>What is your spirit animal?</h3>
+            </div>
+            <label>
+              <div>
+                <input className="inputField" type="text"
+                  value={animal}
+                  onChange={(event) => setAnimal(event.target.value)} required />
```
ericaglimsholt: Write
```
<input 
   className="inputField" 
   type="text"
   value={animal}
   onChange={(event) => setAnimal(event.target.value)} 
   required 
/>
```
Or 
```
<input className="inputField" type="text" value={animal} onChange={(event) => setAnimal(event.target.value)}  required />
```
One of them instead of both

code/src/App.js
```diff
@@ -1,9 +1,130 @@
...
+                <span className="fakeRadio" />
+                Yes
+              </label>
+
+              <label>
+                <input
+                  type="radio"
+                  value="no fun"
+                  onChange={event => setFun(event.target.value)}
+                  checked={fun === "no fun"}
```
ericaglimsholt: So if you have it as bool this will look better
```suggestion
                  checked={fun === false}
```

PR #10, opened 2019-12-01
Project 10 - Fredrik Tibbling

codebyfredrik:
https://youthful-mayer-023937.netlify.com/

I'm quite happy with the end result and the overall structure of the project. Focused more on the React/Javascript/Logic part in this assignment. When I get some free time (probably Christmas break) I will go over the design once more. Maybe I will try a Typeform clone or something.

Early on I decided to store all my survey answers in one state object. By doing so I can easily add more questions without having to create new logic to handle additional answers from the user.

I decided to created some validation for the input text fields. Right now the validation rule states that the input must be minimum 3 and maximum 25 characters. If your answer is validated correctly a green border is conditionally rendered around the input field, otherwise it's rendered red with an error message. 

I also implemented a character counter that is updated on each key stroke as well as a progress bar component to indicate at what step you are in the survey.

I'd like to have general feedback on project. 

I would also like to know how I can render the default border color (grey) of the input field (first name and last name is currently green because it's conditionally rendered based on an error prop) when the page is loaded the first time and the input field has not been in focus (see Input.js).

The button to submit the form is conditionally rendered based on the two text input fields. If the two fields have been in focus once the error state is set correctly and the submit button is correctly disabled. But if the two fields have not been in focus the submit button will not be disabled. How can I solve this?

### DIFF
--- a/code/src/components/form/Form.js
+++ b/code/src/components/form/Form.js
@@ -1,0 +1,96 @@
+import React, { useState } from 'react';
+
+export const Form = () => {
+  const [input, setInput] = useState({});
+
+  const handleInputChange = e => {
+    const target = e.currentTarget;
+
+    setInput({
+      ...input,
+      [target.name]: target.value
+    });
+  };
+
+  const handleSubmit = e => {
+    e.preventDefault();
+    console.log(input);
+  };
+
+  const handleResetForm = () => {
+    setInput({});
+  };
+
+  return (
+    <div>
+      <p>First name: {input.firstName}</p>
+      <p>First name: {input.lastName}</p>
+
+      <form onSubmit={handleSubmit}>
+        <label>First name:</label>
+        <input
+          name="firstName"
+          value={input.firstName || ''}
+          type="text"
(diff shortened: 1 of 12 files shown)


code/src/components/form/Form.js
```diff
@@ -0,0 +1,96 @@
...
+          <option value="Angular/Angular.js">Angular/Angular.js</option>
+          <option value="ASP.NET">ASP.NET</option>
+          <option value="Express">Express</option>
+          <option value="Spring">Spring</option>
+          <option value="Vue.js">Vue.js</option>
+          <option value="Django">Django</option>
+          <option value="Flask">Flask</option>
+          <option value="Laravel">Laravel</option>
+          <option value="Ruby on Rails">Ruby on Rails</option>
+          <option value="Drupal">Drupal</option>
```
ericaglimsholt: For this to be less code you can add the values as an array and then use `map()` to get the different `<option>`. You can read about how to do it here: https://reactjs.org/docs/lists-and-keys.html

PR #11, opened 2019-12-01
Project-Survey - Susanna Sjöberg

SanSjo:
Link to my deployed project: https://project-survey.netlify.com/

This week was fun, i'm not entirely sure if i understand how everything with state and useState works together with the inputs but it will hopefully sink in eventually, like with everything else:)

I had some struggles to get my checkboxes to be multi checked. And how to get the summary text to replace the form when submit, but I got that figured out after wednesdays codealong.

Not what I can think of now.


code/src/index.css
```diff
@@ -5,9 +5,96 @@ body {
     sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background: rgb(230, 223, 223);
+  background-image: url("./assets/hero-image.jpg");
+  background-size: cover;
```
ericaglimsholt: To make the image to be covered in the whole page you need to add a height `height: 100vh;`

code/src/App.js
```diff
@@ -1,9 +1,138 @@
-import React from 'react'
+import React, { useState } from "react";
+import "./index.css";
+import { Summary } from "Summary";
 
 export const App = () => {
+  const sizeComp = ["0-10 people", "10-50 people", "50+ people"];
```
ericaglimsholt: If all have "people" after you can add it to the output
```suggestion
  const ages = ["0-10", "10-50", "50+"];
```

code/src/Summary.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import "./Summary.css";
+
+export const Summary = props => {
+  console.log("props in summary:", props);
+
+  return (
+    <div className="summaryContainer">
+      <div className="SummaryStyle">
```
ericaglimsholt:
```suggestion
      <div className="summaryStyle">
```

PR #12, opened 2019-12-01
Survey - Aleksandra

AleksandraMuller:
Link to my deployed project:

https://survey-by-aleksandra.netlify.com/

This week felt good. I learned a lot about state and find it logical. I started to code the project from the videos that we got this week on Canvas. I struggled the most with hiding the form. 

I started with some strech goals this week as well. I styled the form and placed some parts of the form in Components to practice props even more.
I also used destructing. Afterwards, I created multi-step form. I figued out how to make it work, though I wonder if that is 
the easiest soultion, with the if statements.

I struggled with the validation a bit as it does not work exactly the same way as in HTML. Finally, I decided to create setError state that that I combined with
the state that is responsible for which question is shown and it worked. 

All content of this week is clear.

Best, Aleksandra


code/src/App.js
```diff
@@ -1,9 +1,145 @@
...
+
+    if (section === "firstQuestion") {
+      setSection("secondQuestion");
+    } else if (section === "secondQuestion") {
+      setSection("thirdQuestion");
+    } else if (section === "thirdQuestion") {
+      setSection("fourthQuestion");
+    }
+
+    if (name.length <= 2) {
```
ericaglimsholt: Hard to take this survey if your name is My for example ;)

code/src/App.js
```diff
@@ -1,9 +1,145 @@
...
+                className={"radio"}
+                onChange={handleRadioChange}
+              />
+              <RadioButton
+                label="No"
+                value="no"
+                checked={quality === "no"}
+                className={"radio"}
+                onChange={handleRadioChange}
+              />
```
ericaglimsholt: Could be set to a bool instead? (true/false)

code/src/App.js
```diff
@@ -1,9 +1,145 @@
...
+              <InputTextarea
+                value={experience}
+                onChange={event => setExperience(event.target.value)}
+                className={"inputTextarea"}
+              />
+              <Submit
+                onClick={handleSubmit}
+                text={"Submit"}
+                className={"btn btn-submit"}
```
ericaglimsholt: How come you are adding brackets to the `className`? :)

PR #13, opened 2019-12-01
Nasim Mahzoun - Project-survey

Nasimmhn:
Link to my deployed project:

https://friends-quiz-by-nasim.netlify.com/

Compare to the last project, this week went smoother, maybe because we already got the hints in the video tools and also code alongs with Damien, which was appreciated. Im still struggling with React components, how to divide them, how should I think when it comes to chunking files. For instans, I created a stylesheet just for my form, but it became a problem with media query. To sum up: I got some help regarding some JS functions, but the more I asked for help the more it became clear. I will definitely go back and do the stretch goal 🤞🏻 

Are there any differences between writing statements, conditionals function in JS and React? 
And also when it comes to accessibility, I couldn't fint out how to enter the answer by using the 'space' on keyboard. This is because I replaced the radio input with a styled span, but I would still wanna know how (if possible).


code/src/Components/Form.js
```diff
@@ -0,0 +1,207 @@
...
+}
+// An object for the correct answers, which are imported to Summary, to compare the user-answer with the currect answer
+const CorrectAnswers = {
+  married: married[3],
+  spots: spots[2],
+  places: places[2],
+  pets: pets[3],
+  favSongs: favSongs[1],
+  reasons: reasons[3]
+}
```
ericaglimsholt: Love this!

PR #15, opened 2019-12-02
Project 10 survey, Angelica Brodin

Angelicabrodin:
Link to my deployed project:
https://thechristmassurvey.netlify.com/

I thought this weeks project was really fun and I also felt that understood the logic in the code, 
which I have not felt for a while when starting with java script(react,  I also asked a lot of questions early on, so I got a lot of help instead of spending hours struggling. So that is something I will do more of. As always, I could have spent more time and added some more styling to the form, but I am overall very content with this weeks project!

I still struggle with knowing how many components you can create and how you mount them. 
like for instance, could every drop down, form or radio button be placed in its own component? But I guess this will come with time and practice. Also I struggled with using map() in the radio buttons instead of writing them over and over in the code, I just could not get the answer (the chosen button) to show when pressing submit, so some examples would be nice :)


code/src/App.js
```diff
@@ -1,9 +1,101 @@
...
+  const [movie, setMovie] = useState("");
+  const [showResult, setShowResult] = useState(false);
+  const [songName, setSongName] = useState("");
+  const candyGroups = ["Gingerbread", "Chocolate", "Butterscotch"];
+  const [candyGroup, setCandyGroup] = useState();
+
+  const handleSubmit = event => {
+    event.preventDefault();
+    setShowResult(true);
+  };
```
ericaglimsholt: Would be nice with some error handeling so it doesn't always get true when you pressed the button (only when you have values)

PR #17, opened 2019-12-02
Project Survey - Petra Brinck 

petrabrinck:
Link to my deployed project: https://modest-hopper-c5f89b.netlify.com

My reflections on how this week's project turned out: Pretty good I think. We started up this week together (me ad Theresa) which felt great. We got stuck when the submit button wouldn't work and that was because we divided the components (like we did with the music project) but Jennie solved it with us. 

I still battle with different turn outs depending on the browser I look in when I deploy so I am confused :-)


code/src/App.js
```diff
@@ -1,9 +1,114 @@
-import React from 'react'
+import React, { useState } from 'react'
+import { Summary } from './Summary'
+
+const adventureGroup = ["surfing", "climbing", "hiking"];
```
```suggestion
const adventureGroups = ["surfing", "climbing", "hiking"];
```

code/src/App.js
```diff
@@ -1,9 +1,114 @@
...
+const destinationGroup = ["Costa Rica", "Mount Everest", "Machu Picchu"];
```
```suggestion
const destinationGroups = ["Costa Rica", "Mount Everest", "Machu Picchu"];
```

code/src/App.js
```diff
@@ -1,9 +1,114 @@
...
   return (
     <div>
-      Find me in src/app.js!
+
+      {!submitted && (
+
+        <form onSubmit={event => event.preventDefault()}>
+
+          <div className="heroHeader">
+            <h2>Plan your 2020 adventure!</h2>
```
```suggestion
            <h1>Plan your 2020 adventure!</h1>
```

code/src/App.js
petrabrinck: Great, thanks! :-)

code/src/App.js
petrabrinck: Great again, thanks! :-)

code/src/App.js
petrabrinck: I started out with H1 but it looked huge on my laptop, but maybe its the best choice when it comes to accessibility?

petrabrinck:
Answers: 
Re: Roboto font as files instead of including it as a link from Google fonts? 
I didn't know where to put it in React and I forgot to ask/double check with someone.   


Also - what do you mean by this?
"I still battle with different turn outs depending on the browser I look in when I deploy so I am confused"

When I look at this project on my laptop, on my Iphone XS, in Safari and in Chrome, the content is showing with different margins so I don't now if I should rely on one (or none) of them? :-/ 
Thank you! 
Petra

PR #18, opened 2019-12-02
Project 10 - Survey, Linda Isell

liissee:
Link to my deployed project:
https://project-survey-linda.netlify.com/

Fun project and I'm starting to understand the logic in React but not entirely.. Last week I felt that I had too many components and this week I have too few..
I started early with the project and I spent more time on the styling this week which was fun.

- Accessibility: How do I get the radio-buttons focused when I tab through the page? 
- Same for the emojis? I have labels around them but I can’t tab between the two?
- in firefox and on the phone the range-bar icon(emoji beer) has a circle around it.. 
- Progressbar,  I found a tutorial how to create a progressbar and I managed to link it with my "continue" and "go back" buttons/functions but there is bug, if you click continue almost to the last question and then "go back", then the progress bar i 100% even if you are at the first question.. whyyyy? :(
-  I have a lot of code for having multi-step questions and I would be so happy for any suggestions on how to make it shorter. 
-  And I still struggle how the components are linked to each other.. I wanted to have the progress bar in a separate component but I couldn't figure out how to use a function that is inside a different component...


code/src/App.js
```diff
@@ -1,9 +1,330 @@
-import React from 'react'
+import React, { useState } from 'react'
+import { Summary } from './Summary'
+import "./app.css";
+import './progressbar.css';
```
ericaglimsholt: Decide if you want the project to have `"` or `'` and stick to it

liissee: Hej Erica,
 Tack! Angående aria-label, kan du ge ett exempel hur/var jag ska använda det? Jag har använt aria-label men isf på fel sätt? Jag har provat att ändra till aria-label="radio-button"> men det funkar ändå inte. 

{question === 6 && (
            <div className="question">
              <div className="feelings">
                <h3>Are you happy with that amount?</h3>
                <div className="emojis">
                  <label>
                    <input type="radio"
                      value="happy"
                      onChange={() => setHappiness("happy")}
                      checked={happiness === "happy"}
                    />
                    **<span role="img" aria-label="happy face">**
                      😅
                </span>
                  </label>
                  <label>
                    <input type="radio"
                      value="sad"
                      onChange={() => setHappiness("sad")}
                      checked={happiness === "sad"}
                    />
                    **<span role="img" aria-label="sad sweaty face">**
                      😓
                </span>
                  </label>
                </div>
              </div>
            </div>
          )}

PR #22, opened 2019-12-02
Survey about Christmas - Matilda Arvidsson

HIPPIEKICK:
**Link to my deployed project:** [https://matilda-arvidsson-project-survey.netlify.com/](https://matilda-arvidsson-project-survey.netlify.com/)

**My reflections on how this week's project turned out:** This was a fun project! A lot of small "wins" all week, and not a lot of time stuck as last week. I'm quite happy with how the page turned out, but I am not as happy with my code. It feels even messier than last week. This week I started doing everything in the same js-document and today I looked at my monster and thought about dividing it into components, but I'm not sure how to do that. Maybe this will be my Christmas break task.

I didn't put much effort to responsiveness, but I really tried to make the page accessible (with autoFocus and connecting Enter to "Next question"-button instead of "Submit"-button etc).

- It took some time before I found out I could use autoFocus for inputs. At first, where ever I read about it, people wrote about ref's and classes and this and that was a little overwhelming. Is this something we will learn another time?

- Also, as I mentioned above, my code is messy. Partly because I have so many divs (with bad names) and partly because I have everything in the same js-file. How many components would make sense this week? One for the whole form and one for summary? Or is it possible (and does it make sense) to have one component per question?

- Naming convention for classNames. I thought, since we're using them in CSS, we should use kebab-case, but I think you guys use camelCase? Why? :) (PS. I know I have two CSS-classes with camelCase-names, and I will change all of them when you tell me I should use camelCase).

- Images as radio-buttons, could I have mapped over them in some way, and saved some space?


code/src/Form.js
```diff
@@ -0,0 +1,300 @@
+import React, { useState } from 'react'
+import tree1 from 'images/tree1.jpeg'
+import tree2 from 'images/tree2.jpeg'
+import tree3 from 'images/tree3.jpeg'
+import tree4 from 'images/tree4.jpeg'
+
+export const Form = () => {
+  const [ageGroup, setAgeGroup] = useState()
+  const [merryChristmas, setMerryChristmas] = useState('')
```
ericaglimsholt: This naming doesn't say so much of what it's representing

code/src/Form.js
```diff
@@ -0,0 +1,300 @@
...
+            }
+
+            {/**QUESTION 3 */}
+            {question === 'christmasDecoration' && (
+              < div className="select-div" >
+                <label>Favourite Christmas decoration:
+                  <select
+                    onKeyPress={(event) => {
+                      if (event.key === 'Enter') { handleContinue() }
+                    }}
```
ericaglimsholt: This is really nice!

PR #24, opened 2019-12-02
Jola-Chabik-project-survey-

ucha3i:
Link to my deployed project:

https://app.netlify.com/sites/ucha3i-project-9-survey/overview

I find this project surprisingly not as complicated as I thought it is at the beginning, though I chose to make a simple version.  I don't want to overcomplicate things for myself right now and just stay on a way where I start to feel like I'm understanding what I'm doing. 

I'm think I'm fine at the moment. But surely, there will be bottoms of the rollercoaster in the future.


code/src/App.js
```diff
@@ -1,9 +1,105 @@
...
+          <h2>Name</h2>
+          <input
+            type='text'
+            onChange={event => setName(event.target.value)}
+            value={name}
+          />
+    
+        <h2>First, pick your style:</h2>
+        
+        {activities.map(group => (
```
JennieDalgren: Nice done with the map

code/src/App.js
```diff
@@ -1,9 +1,105 @@
...
-  )
+    </section> 
+  
+    )}
+
+  {showResult && (
+        <p className='result'>
+          Hi {name}! You're a true adventurer! You can't wait to go to {place} and bring a fresh {activity} adventure to this season! But hold you horses. You still have to wait for a bit more snow! Relax yourself with some great <a href="https://www.mountainsonstage.com/">mountain movies </a> and {food} till then.
+        </p>
+      )}
```
JennieDalgren: I really liked this content wise :) Super nice.

One thing you could have done here is to create a component out of the summary/result. just to practise splitting it up a bit.

ucha3i:
Will do that, also because I'd like to manage three different result and see where I'll go with that in code :) Thank you very much for the review!

PR #25, opened 2019-12-02
Survey

camillalof:
Link to my deployed project:

I couldnt see the final page when I deployed it and I think it have something to do with that I tried to merge my components-branch onto the masterbranch. I didnt remember how to do it so I googled it and it seemed fine. It have happened to me that the page dont show until later (?) so here is the adress just in case it might work: https://cocky-hugle-3a8637.netlify.com/ but if not can you please help me to se over the branches. I think the problem is there if there is a problem.

Ugly but functional. :) No but I am proud that I could divide the page into different components and understand what I was doing.

I have problems with understanding hov to make flexbox work in react. Therefore my page is ugly. And I could not get the dropdown value to be connected to the summary.

Can you please look at my code and see what I did wrong when mearching the components-branch?

Can you please help me understand why the page is not showing the value from the dropdown?

Can you please help me explain where to put the divs if I want to create flex-children in react. Lets say for example that I want to have the whole form (buttons and all centered centered, with ceneter center I meen like the text is in in my summary in the form)?

I hope its okey that I ask this! Maybee its easy things for others, but I didnt understand. Thanks alot! :)

/Camilla

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,16 @@
 import React, { useState } from 'react'
+import { Summary } from './Summary.js'
+import { Happiness } from './Happiness'
+import { Input } from './Input'
+import { Dropdown } from './Dropdown'
 
 export const App = () => {
-  const [happiness, setHappiness] = useState('happy');
-  const [feeling, setFeeling] = useState('');
+  const [happiness, setHappiness] = useState('')
+  const [feeling, setFeeling] = useState('')
+  const [today, setToday] = useState('') //används inte
   const [showResult, setShowResult] = useState(false)
+  const [question, setQuestion] = useState('feeling')
+  const [location, setLocation] = useState('');
   
 
   const handleSubmit = (event) => {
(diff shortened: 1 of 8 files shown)


code/src/App.js
```diff
@@ -1,9 +1,85 @@
...
-    <div>
-      Find me in src/app.js!
-    </div>
+  <div>
+    {showResult && (
+      <Summary feeling={feeling} happiness={happiness} />
+    )}
+
+    {!showResult && (
+      <form onSubmit={handleSubmit}>
```
JennieDalgren: This `handleSubmit` happens after the first question, if you press the enter key. (which is the same as submitting your form).
See my suggested change above on line 15-26

code/src/App.js
```diff
@@ -1,9 +1,85 @@
...
+
+  const handleContinueClick = () => {
+    console.log('click continue')
+    if (question === 'feeling') {
+      setQuestion('happiness') 
+    } else if (question === 'happiness') {
+        setQuestion('treatment')
+      
+    }
+  }
```
JennieDalgren: Since it's possible to submit the form already on question 1 when using the enter key on your keyboard you could bake this into one function.
something like:
```
 const handleSubmitOrContinue = (event) => {
   event.preventDefault()

    if (question === 'feeling') {
      setQuestion('happiness') 
    } else if (question === 'happiness') {
        setQuestion('treatment')
    } else {
       // do the actualy submit.
       setShowResult(true)
    }
  }
```

code/src/App.js
```diff
@@ -1,9 +1,85 @@
...
+
+      {question === 'feeling' && (
+        <div className="question">
+          
+          <h1>Fill in your name:</h1>
+        
+          <Input 
+          label="Name"
+          value={feeling}
+          setValue={setFeeling} />
```
JennieDalgren: Maybe this should be called `name` and `setName`??

PR #28, opened 2019-12-02
Week 10 handin

tehvicke:
Link to my deployed project: https://zen-noether-4db04f.netlify.com/

Interesting! With a new framework my previous knowledge is not that helpful, and I've learned quite a lot both regarding React and how to structure code in a good way.

Nothing in particular.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,117 +1,47 @@
 import React, { useState } from "react";
-// import SerializedForm from "./SerializedForm";
+import Form from 'Form.js';
 
 export const App = () => {
-  const [isAnswered, setAnswered] = useState(false);
+  const [isAnswered, setAnswered] = useState(false); /* if Answered (submitted) then show answers */
 
-  return (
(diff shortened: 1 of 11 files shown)


code/src/App.js
```diff
@@ -1,9 +1,29 @@
-import React from 'react'
+import React, { useState } from "react";
+import Form from "Form.js";
+import Answers from "Answers.js"
```
ericaglimsholt: Tip is to add these to a `Components` folder so it will be easier to navigate if you add 50 more files to the project later on

code/src/InputTypes/RadioInput.js
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+
+const RadioInput = ({ label, checked, onChange }) => {  
+  return (
+    <label className="radio-label">
+      <input
```
ericaglimsholt: Maybe add the input as `required`?

PR #30, opened 2019-12-02
rsvp-survey Cislowski

KarolCislowski:
Link to my deployed project:
https://rsvp-cislowski.netlify.com/

Very nice project, which can be easily adapt to many other projects :) 

I'm all good :)


code/src/containers/Survey/Survey.js
```diff
@@ -0,0 +1,74 @@
...
+          onNextSite={nextSiteHandler}
+          title={data.title}
+          place={data.place}
+          date={data.date} />
+      }
+      {questions}
+      {
+        siteIndex === questions.length && <Summary answers={answers} />
+      }
+      <ProgressBar actual={siteIndex} final={questions.length} />
```
JennieDalgren: Minor thing, but maybe this shouldn't be shown on the first Welcome screen?
Fix that by something like `{siteIndex > -1 && <ProgressBar />}`

code/src/containers/Survey/data.json
```diff
@@ -0,0 +1,27 @@
+{
```
JennieDalgren: Nice to extract the data into its own file.

code/src/containers/Survey/Text/Text.js
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from 'react'
+
```
JennieDalgren: really nice that you extracted the inputs to reuse them. Although, should allergens actually be required? Maybe you could pass required as a prop instead? To make it even more modular?

PR #31, opened 2019-12-03
.

horsefacegirl:
Link to my deployed project:
https://pedantic-jones-7d3652.netlify.com

I would like to spend some more time on styling, not really satisfied with that part. I could also divide the app into Components.


horsefacegirl:
> Good work on your project!
> I'm interested in knowing how come you have named your files as `cat.icon.svg`? Never seen it before :)

Haha I don't know actually! Will probably change that aswell if I pick up this projoct again to refine it.

PR #37, opened 2019-12-17
Survey-Ivett

Lilivet:
Link to my deployed project:
https://ivett-survey.netlify.com/

I had to many things on my private and work life that I could not focus on this project. It took me forever to finish it. I would like to change the style more, but it will have to wait until Christmas vacations

It took me a while to understand how to relate the components with the main App component. I still struggle with map and props...I think it is a matter of continuing using them.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -44,7 +44,7 @@ export const App = () => {
     <div>
       {showResult && (
         <div>
-          <h2>Thank you for participating in this survey! We capture your suggestions:</h2>
+          <h2>Thank you for participating in this survey! We captured your suggestions:</h2>
           <p>You would like to have the AW <span>{awFrequency}</span>.</p>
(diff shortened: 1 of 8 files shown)


code/src/App.js
```diff
@@ -1,9 +1,96 @@
...
 
+// Set of arrays
+const awFrequencies = ['Every week', 'Every two weeks', 'Every month']
+const themeFrequencies = ['Every afterwork', 'Every second afterwoks', 'once a year']
+const themeAwChoices = ['YES', 'NO']
+
+// Hooks that I am using in the App
 export const App = () => {
+  const [awFrequency, setAwFrequency] = useState(awFrequencies[0])
+  const [themeAw, setThemeAw] = useState('')
```
ericaglimsholt: What about making this to a bool instead? Just because the answers is yes or no only

PR #47, opened 2020-03-18
Survey form react Hanna Rüter

Ruterryter:
Link to my deployed project:
https://elegant-meninsky-46139a.netlify.com/

It was a fun project. If I had more time I would go for the stretch goals.


code/src/App.js
```diff
@@ -1,9 +1,89 @@
...
+import './index.css'
+
+
+const ageGroup = ['too young', '19-35', '35-65', '+65'];
+const option0 = ''
+const option1 = 'Candlelight dinner in a cosy restaurant'
+const option2 = 'Netlfix and Chill'
+const option3 = 'Dancing on my own'
+const option4 = 'Party with my friends'
+const option5 = 'Headbangers ball concert'
```
ericaglimsholt: Maybe this would suit as an object or array instead?

code/src/App.js
```diff
@@ -1,9 +1,89 @@
...
+                />
+                {group}
+              </label>
+            ))}
+
+            <h2> What is a perfect evenig according to you?</h2>
+            <select className="hobby"
+              onChange={event => setHobby(event.target.value)}
+              value={hobby}>
+              <option value={option0}>~ Select option ~</option>
```
ericaglimsholt: I guess you don't want this as an option? :)
```suggestion
              <option value={option0} disabled>~ Select option ~</option>
```

code/src/App.js
Ruterryter: Thanks for the review! I wanted to do something with this but was unsure how to make it in another way, would you please give me a hint on how to change this?.

PR #57, opened 2020-03-19
Survey-project Emma Prené

emmapre:
Link to my deployed project:
plant-survey.netlify.com
Quite well in the end. I tried to make more components without making everything work first so I changed back to keeping the form in one piece. I couldn't figure out to connect the components and their states. 
How the ! Summary ? part really works.


code/src/App.js
```diff
@@ -1,9 +1,70 @@
...
+      {!showSummary ? (
+        <form onSubmit={handleSubmit} className='box form'>
+          <div>
+            <label>
+              <h3>How do you live?</h3>
+              <select
+                onChange={event => setLivingPlace(event.target.value)}
+                value={livingPlace}
+                required>
+                <option value=''>Select housing</option>
```
ericaglimsholt: Guess you don't want this to be a noption
```suggestion
                <option value='' disabled>Select housing</option>
```

PR #59, opened 2020-03-20
Project Survey Lisa Sturegård

LisaStur:
Link to my deployed project:
https://quizzical-bassi-3df6ac.netlify.com/

Very good project, learnt a lot! The whole setup between different components and files, with props and import-export.


code/src/components/HoardingPlaceEntry.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const HoardingPlaceEntry = ({place, setPlace}) => {
+  return (
+    <form className="hoarding-place">
+      <h2>Where do you prefer to stock up?</h2>
+      <select className="hoarding-select"
+        onChange={event => setPlace(event.target.value)}
+        value={place} 
+      >
+        <option value=''>Hoarding Place:</option>
```
ericaglimsholt: Guess you don't want this as an option
```suggestion
        <option value='' disabled>Hoarding Place:</option>
```

LisaStur: Ah, big thanks, very useful feedback!

PR #64, opened 2020-03-20
Project survey - Johanna Rådén

johannaraden:
Link to my deployed project:
https://modest-wilson-211340.netlify.com/

I am quite happy with the result. In terms of CSS I would've like to have time to do some more. I mostly focused on getting the functionality to work. 

I would like to know more about what I can do with props. I did not know before going in to this week that I could tranform the props into array values in order to use them in for example a map method.

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,105 @@
-import React from 'react'
+import React, { useState } from "react";
+import { Summary } from "./Summary";
+import "./App.css"
 
 export const App = () => {
+  const animalsArray = ["Cat", "Dog", "Bunny", "Parrot"];
+  const [gender, setGender] = useState("");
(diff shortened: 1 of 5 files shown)


code/src/App.js
```diff
@@ -1,9 +1,105 @@
...
+                />
+            </section>
+            {/* Choose input type number since the browser support for the type="tel" is very low according to w3School. At least now it gives an error if you would type in letters instead. */}
+            <section>
+              <label>Gender</label>
+              <select
+                id="gender"
+                onChange={event => setGender(event.target.value)}
+              >
+                <option value="">Select </option>
```
ericaglimsholt: I guess you don't want this to be an option!
```suggestion
                <option value="" disabled>Select </option>
```

PR #70, opened 2020-03-21
Form by Lisa

L-Lisa:
Link to my deployed project: https://confident-roentgen-a7ddea.netlify.com/

My reflections on how this week's project turned out: I first made a basic form and then when I got the hang of it started again and did this one, time was a little short so had to cut design. I'm pretty pleased that I   f i n a l l y  figured out how to use the useState hook in the different sections.

The hover effect on the submit button disappears sometimes and I really do not understand why, if you can see that please do let me know. Besides that, I'm happy for any feedback on how to make my code and understanding of React better.

My assignment was handed in on Canvas yesterday. I first handed in another form but then couldn't create a pull request from there so handing this one in, I hope that is ok.

Have a great day
Lisa


code/src/OrderSize.js
```diff
@@ -0,0 +1,31 @@
...
+    )}
+
+
+
+
+
+
+
+
+
```
ericaglimsholt:
```suggestion
```

PR #86, opened 2020-10-22
Project Survey - Vanessa Sue Smith

VanessaSue27:
Link to my deployed project:
https://serene-engelbart-f399ff.netlify.app/

This was a very challenging project, mainly because I started with this idea of making my survey dynamic and that became more complex than I thought, especially since it had to be built with React, which we are quite new to. I listed some steps on how I built it on the ReadMe file.
I tried my best and made it work as intended at the end, but I still have several doubts about best practices.
I feel that I was much more comfortable with JavaScript, so becoming friends with React is still a work in progress. I think I just need to get to the point where I understand its power and why it is such an effective tool to use 😝

I need a lot more explanation on how to use State especially when I have a components tree like I did on this project (5 different components). I connected the App component and the first question through State passed as props, but from the second question to the final result, they have local state and the data entered by the user is passed via props until the final summary text is generated.
It works and it's a start, but would like to know more about best practices on how to manage States.


Snabbina (approved):
Hey Vanessa🍉! 

🏆 Great job, your code is really structured and the naming of components and classes easy to follow. Nice, clean looking layout with the landing page and the framed questions taking you step by step through the survey!

I have focused on testing out the survey from a user point of view and will give some comments from a picky users-perspective🤓.

Hope the comments can add some extra perspective to your already great project! /S

As I see  you have passed the initial requirements so you are approved (by me 😊)

code/src/App.js
```diff
@@ -1,9 +1,54 @@
...
+import { FirstQuestion } from './components/FirstQuestion';
+import 'components-css/landing-page.css';
 
 export const App = () => {
+  const [visible, setVisible] = useState(false);
+  const [destinationOption, setDestinationOption] = useState();
+
+  const checkInput = () => {
+    setVisible(true);
+  };
```
HIPPIEKICK, line 11: You could either change the name of this function to `startSurvey` or you could just change the state directly in the onclick. Eitherway - there's no input checking by clicking a button, right? ;)

code/src/App.js
```diff
@@ -1,9 +1,54 @@
...
+    setDestinationOption(newDestination);
+  };
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <>
```
HIPPIEKICK, line 18: No need for a React.fragment when the `main` is already wrapping it all

code/src/components/FinalText.js
```diff
@@ -0,0 +1,23 @@
...
+    <section className="final-text">
+      <div className="final-result-container" id="final-text">
+        <img className="final-img" src="assets/final-img.png" alt="Group of friends camping" />
+        <p className="dream-text" tabIndex="0">On your dream vacation </p>
+        <p className="result-text" tabIndex="0">
+          ...you&apos;ll find yourself chillin&apos; at
+          the {props.resultDestination} with {props.resultNumber} of
+          your best friends, drinking cocktails during the {props.resultDay}.
+        </p>
+        <button className="restart-button" onClick={restart} type="button">Start again!</button>
```
HIPPIEKICK, line 19: Tried this: `location.reload()`?

code/src/components/FinalText.js
```diff
@@ -0,0 +1,23 @@
...
+      </div>
+    </section>
+  )
```
HIPPIEKICK, line 22: ;

code/src/App.js
```diff
@@ -1,9 +1,54 @@
...
+            <a href="#first-question" aria-hidden="true" tabIndex="-1">
+              START
+            </a>
+          </button>
+          <img
+            src="assets/landing-page-img.png"
+            className="img-landing"
+            alt="Vacation planner" />
+        </section>
+        {visible === true && (
```
HIPPIEKICK, line 41: `{visible && (` is enough

HIPPIEKICK (commented):
Hi Vanessa!
Really cool that you did your own thing with this form - super good practice to really break it down and see what works or not!

Regarding your structure, perhaps it would make more sense to mount all questions in the App component or in a `Survey.js` or `Form.js` than to have the `FirstQuestion.js` component holding all of them. You could for example change your visible state to a state called `section` or `question` with a starting state of `0` and then increment that state in every question component. And then mount them conditionally like:
```
{section === 1 && <FirstQuestion/>}
{section === 2 && <SecondQuestion/>}
```

That would also allow you to hide the former components when you mount a new one (e.g. hide the form when you show the summary).

Also, if you for some reason would want to change the order of your questions it would be easier if they are all mounted in the same component, instead of needing to change inside several components. Hope that makes sense! :)

PR #108, opened 2020-10-25
project-survey-liza-minican

liza-minican:
Link to my deployed project: https://coffeesurvey-liza-minican.netlify.app/

My reflections on how this week's project turned out: I spent much more time on it than I though I would, the project was interesting and I learnt a lot about components. I need a lot of practice to truly understand the states. 

Things I'd like to have clarified or explained in more detail: I would like to clarify how to use output in range input and I want to understand how not to show some elements when you display a certain component.


code/src/App.js
```diff
@@ -1,9 +1,10 @@
...
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
+    <>
+      <Form />
+    </>
+  );
 }
```
carolinebirgersson, line 10: Clean code! I like that you have made components out of each question and therefor are able to just insert the form here.

code/src/css components/button1.css
```diff
@@ -0,0 +1,16 @@
...
+  width: 100px;
+  font: bold 15px arial, sans-serif;
+  text-shadow: none;
+}
+.submitbutton:hover {
+  background: rgb(248, 196, 204);
+}
+.submitbutton:focus {
+  outline: thin dotted;
+}
```
carolinebirgersson, line 16: My suggestion here is to make 1  CSS-file for "Buttons" and link the same file to each component that uses it. Then you don't have to have repeated code in many files – which makes it hard to control if you want to make changes.

code/src/index.css
```diff
@@ -5,9 +5,39 @@ body {
     sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background-image: url("https://res.cloudinary.com/dgujdyxlv/image/upload/v1603403593/pexels-lukas-942803_qklgk6.jpg");
+  background-repeat: no-repeat;
+  color: white;
 }
```
carolinebirgersson, line 12: I really like your background with the survey on top of it! Good job!

code/src/js components/Button1.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import "css components/button1.css";
+
+export const Button1 = () => {
```
carolinebirgersson, line 4: I think it's a bit vulnerably to name them Button1 etc. Maybe not in this case when the site is shallow but in a major project where it's a lot of buttons it can be hard to remember which button belongs to which name of it. Then it might be better to call them submitButton etc, like you do in the classname :) But this wouldn't be a problem if you gathered them in one file like I suggested before.

code/src/js components/Button3.js
```diff
@@ -0,0 +1,15 @@
...
+
+const goback = () => {
+    document.location.href = "";
+};
+
+export const Button3 = () => {
+  
+  return (
+    <>
+    <button className="resetbutton" onClick={goback}type="reseting">reset</button>
```
carolinebirgersson, line 12: I like that you uses onClick here instead of adding an a-element. You know your Javascript-skills! ;)

code/src/js components/Form.js
```diff
@@ -0,0 +1,77 @@
...
+import { Questiondrink } from 'js components/Questiondrink';
+import { Questionamountofcoffee } from 'js components/Questionamountofcoffee';
+import { Questionrange } from 'js components/Questionrange';
+import { Button1 } from 'js components/Button1';
+import { Header } from 'js components/Header';
+import { Summary } from 'js components/Summary';
+
+export const Form = () => {
+  const [answername, setAnswername] = useState ("");
+const [answeraddress, setAnsweraddress] = useState ("");
```
carolinebirgersson, line 13: I think it should be an indentation on all of these that doesn't have it?

code/src/js components/Form.js
```diff
@@ -0,0 +1,77 @@
...
+        answeramount={answeramount}
+        onAmountChange={handleAmountChange}
+    />
+    <Questionrange
+        coffeeprice={coffeeprice}
+        onRangeChange = {handleCoffeepriceChange}
+        />
+    <Button1 />
+  </form>
+  ) : ( <Summary name = {answername}
```
carolinebirgersson, line 68: Move down {answername}, like you have done on the questions :)

code/src/js components/Questionaddress.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+export const Questionaddress = ({answeraddress, onAddressChange}) => {
+
+return(
+  <>
+      <div>
+        <label tabIndex='0' htmlFor="answerAdress">What is your address?</label>
+  <input
```
carolinebirgersson, line 8: Be aware of the indentation. I think input should be "one in" from label.

code/src/js components/Questionamountofcoffee.js
```diff
@@ -0,0 +1,34 @@
...
+        <p tabIndex='0' >How many cups of coffee do you drink per week?</p>
+            <label tabIndex='0' htmlFor="answerAmount">1-3</label>
+                <input type="radio"
+                onChange = {event => onAmountChange(event.target.value)}
+                value = "1-3"
+                name="cups"
+                checked = {answeramount === "1-3"}
+                required
+                aria-label="Select the amount of cups of coffee you drink per week"
+            />
```
carolinebirgersson, line 16: Try to be consistent with the first "value" after the element name. Sometimes you keep it in the same line as the element (like you've done here). Sometimes you put in on a new row (like you've done on the select-element in Questiondrink.js). I think that you should put it in a new row all the time.

code/src/js components/Questiondrink.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+export const Questiondrink= ({answerdrink, onDrinkChange}) => {
+return(
+  <>
+  <p>{answerdrink}</p>
+      <div>
+        <label tabIndex='0' htmlFor="answerdrink">What is your favourite coffee?</label>
+          <select 
+            name="coffeetypes"
```
carolinebirgersson, line 10: Here's a good example of how you write the indentation. You have a new line for the first value after select (name="cofeetype") – which is one indentation in from the select element. Good job!

code/src/js components/Summary.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import {Button2} from "js components/Button2";
+import {Button3} from "js components/Button3";
+
+export const Summary = (props) => {
+return (
+<>
+  <section className="summary">
+      <h1>Hey {props.name}! We are going to deliver your {props.drink} tomorrow to {props.address}! Enjoy! If you would like to get your coffee delivery {props.answeramount} times per week, please subscribe.</h1>
```
carolinebirgersson, line 9: My recommendation here is to have the text as a` <p>`. Because you only use h1 for headings, and this is more like text-content. And then style them separate in the Header.css as h1 and p instead of baking them together under the header-class :)

code/src/js components/Summary.js
```diff
@@ -0,0 +1,15 @@
...
+    <Button2 />
```
carolinebirgersson, line 10: Get this button (button2) to aline with the button under!

carolinebirgersson (commented):
Good job Liza! I really like your styling and hadn't much to comment on the css, just a suggestion to add the buttons into one file, just to have the code for them minimilized and the possibility to reach them easily if you want to change something that will contain all of them! 

The other comments was about indentation + changing the h1 summary text to a p-element. 

Be proud!

code/src/js components/Questionrange.js
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from 'react';
+import "css components/questionrange.css";
+
+export const Questionrange = ({coffeeprice, onRangeChange}) => {
+
+  return (
```
carolinebirgersson, line 6: One suggestion is to add a label here, just to improve the user experience ;) Like a scale from 1-5 or something.

code/src/App.js
```diff
@@ -1,9 +1,10 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
+    <>
+      <Form />
+    </>
```
EmOlsz, line 8: No need to wrap `<Form />` element inside React Fragment (`</>`) since you return only one element. This is enough :

```
return (
    <Form />
);
```

code/src/App.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
+import { Form } from "./js components/Form"
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';
import ReactDOM from 'react-dom';

import MyComponent1 from './components/MyComponent1';
import MyComponent2 from './components/MyComponent2';
import MyComponent3 from './components/MyComponent3';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/js components/ButtonReset.js
```diff
@@ -0,0 +1,15 @@
...
+const goback = () => {
+    document.location.href = "";
+};
+
+export const ButtonReset = () => {
+  
+  return (
+    <>
+    <button className="button btn-reset" onClick={goback}type="reseting">reset</button>
+    </>
```
EmOlsz, line 13: Same here about wrapping in React Fragment

code/src/js components/Questionamountofcoffee.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+export const Questionamountofcoffee = ({answeramount, onAmountChange}) => {
+return(
+<>
+    <div>
+        <p tabIndex='0' >How many cups of coffee do you drink per week?</p>
+            <label tabIndex='0' htmlFor="answerAmount">1-3</label>
+                <input type="radio"
```
EmOlsz, line 9: Hmm, your inputs are not connected to labels - I tried to click on labels and wasn't redirected.
If you add `htmlFor` attribute to label, you need to remember to add `id` with the same value to input element ;)

EmOlsz (commented):
Hey Liza! Some things needs to be chaged. Take a look on both Caroline's and my comments and provide changes! After that, I'll approve your project.

Code to fix :

- Indentations
- React Fragment
- HTML input and label connection

liza-minican:
Hello,

I have fixed the issues, please take a look.
Thank you!

PR #109, opened 2020-10-25
Week 10 - Project Survey 

Andrea-Osmar:
Link to my deployed project: https://adventureworldtravel.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,160 @@
...
+}
+/* INPUT */
+
+.input {
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+}
+
+input {
```
emeliesv, line 79: Perhaps next time you could style the input:invalid a bit? At the moment, since you have a very dark background, it was a little bit hard to see when the input was invalid with the default styling. Just a tip for the future 🥰

emeliesv (approved):
Very nice work Andrea!
I really liked your survey, it was beautifully styled and written in good code! 😄 

I noticed that you could move on to the next question without selecting/writing anything in your questions, and I don't understand why the 'required'-attribute doesn't work. I think it has something to do with the connection to onSubmit for the form. I did the same thing and was curious to see if maybe the codecoaches has an answer to this? However, if you wanted to do some more validation and make sure that the user writes something, you could make sure that the button is disabled as long as the user input doesn't exist by adding:
`<button disabled={!interest}>` for example in your Radiobuttons.js. 
Your project fulfills all of the general and blue requirements, and some red (validation) and black (one question at a time) requirements as well as far as I can see. I tabbed my way around you page (even invited Daniel 😉) and ran it through Lighthouse and it works great (you could perhaps change the image to a smaller size for optimizing performance in mobile?). 

Nice job! 🥇 

Keep up the good work!

code/src/index.css
```diff
@@ -1,13 +1,160 @@
...
+  border-radius: 3px;
+  background-color: white;
```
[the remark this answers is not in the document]
Andrea-Osmar: Will change that ;) got lazy...

code/src/App.js
```diff
@@ -1,9 +1,103 @@
...
+
+  const handleSubmit = event => {
+    event.preventDefault()
+  }
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
```
EmOlsz, line 22: No need to use React Fragment if you return only one element (`form`) from your component

PR #113, opened 2020-10-25
Week 10  Project Survey  - Karolin Albinsson

karolinalbinsson:
Link to my deployed project:
https://karolin-birthday-invitation.netlify.app/

I think it turned out good. It was fun learning how to use state.


code/src/SubmitButton.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const SubmitButton = () => {
+  return (
+    <button className="submit-button" type="submit">
+      Submit my answers
+    </button>
+  )
```
the-j-curl, line 8: To make this component more usable you could have sent props and then used these props to set a different class, different button type, different text etc, and then re-used it for each question.

code/src/TextInput.js
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div className="question-container">
+      <label htmlFor="textInput" className="question-label">{props.text}</label>
+      <input
+        id="textInput"
+        type="text"
+        onChange={handleChange}
+        value={name}
+        placeholder="Type your name here"
+        required
```
the-j-curl, line 40: Did you manage to get "required" to work?
I had to remove it in my code because it would only execute onSubmit.

code/src/Welcome.js
```diff
@@ -0,0 +1,25 @@
...
+    props.onChange(surveyStarted);
+  }
+
+  return (
+    <section className="form-welcome-message">
+      <h1 className="welcome-header">You have been invited to my {props.data.eventTitle}</h1>
+      <p>It will take place on {props.data.date} in {props.data.location}</p>
+      <p>If you want to attend, please answer these questions. See you!</p>
+      <button
+        tabIndex="0"
```
the-j-curl, line 17: tabIndex is not necessary on the button tag. This tag is focusable by default.

code/src/data.json
```diff
@@ -0,0 +1,27 @@
...
+      "inputType": "Select",
+      "options": ["Hamburgers", "Pizza", "Tacos", "Salad"]
+    },
+    {
+      "questionText": "You will receive a small party gift - pick your favourite color!",
+      "inputType": "Color",
+      "options": []
+    }
+  ]
+}
```
the-j-curl, line 27: Great idea to include this!

code/src/index.css
```diff
@@ -1,13 +1,231 @@
...
+	background-color: transparent;
+}
+
+progress::-webkit-progress-value {
+	background-color: #fff;
+	opacity: 0.3;
+}
+
+output {
+	font-size: 12px;
```
the-j-curl, line 76: It is worth considering that a larger font size on desktop will be easier to read for visually impaired users. (Recommended 16px and above)

the-j-curl (approved):
Wow, I am really impressed with your solution to this project. Using a json file was especially a nice touch. I also love the way the button only appears once the condition is met. 
One improvement to this would be to add a back button to be able to change a previous answer. As a stretch goal.

I noticed that the placeholder text in the input field was slightly hard to read due to a light in the background image. One solution could be to add linear-gradient to your background image to make it slightly darker. I have also added a few other pointers in my comments below.

Overall your code looks super neat and I have really enjoyed reviewing it and learning more about your approach. - I am looking forward to the party! :)

.gitignore
```diff
@@ -1,18 +1,18 @@
...
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+node_modules
+
+# production
+build
+
+# misc
+.DS_Store
```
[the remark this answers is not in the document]
EmOlsz: This file basically says what other files/directories shouldn't be included in github remote repository

code/src/Colorpicker.js
```diff
@@ -0,0 +1,32 @@
...
+        type="color"
+        id="color"
+        name="favcolor"
+        value={color}
+        onChange={handleChange} />
+
+      <button
+        className="next-button"
+        type="button"
+        onClick={() => clickNextButton()}>
```
EmOlsz, line 27: Since you are not passing any arguments inside `clickNextButton()` function, we can simplify this line of code :

```
onClick={clickNextButton}>
```

What's the difference? With your approach, you are creating new function each time you click on button.
With my approach however we are creating function only one time, so we are more efficient

code/src/Form.js
```diff
@@ -0,0 +1,91 @@
+import React, { useState } from 'react';
+import { Colorpicker } from 'Colorpicker';
+import { RadioGroup } from 'RadioGroup';
+import { Select } from 'Select';
+import { Welcome } from 'Welcome';
+import { ProgressBar } from 'ProgressBar';
```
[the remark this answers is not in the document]
EmOlsz: And also one empty line between out and local dependencies would be nice

code/src/ProgressBar.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export const ProgressBar = (props) => {
+  return (
```
EmOlsz, line 4: Well done :)

code/src/Welcome.js
EmOlsz: 👍

PR #114, opened 2020-10-25
Sandrine's weekly covid-19 evaluation

MindstormingAB:
Link to my deployed project: https://weekly-evaluation.netlify.app/

My reflections on how this week's project turned out: I didn't have so much time to spend on this project so I mainly focused on the blue requirements.

Things I'd like to have clarified or explained in more detail: I was not able to add a footer as a React component that will behave in a correct way for all screen sizes.


code/README.md
```diff
@@ -1,19 +1,25 @@
...
+The first question is displayed when the Start button is clicked.
+The following questions are displayed when the previous one is answered.
+When the last question is answered, the button Submit is revealed.
+By clicking on Submit, a summary is displayed instead of the survey form.
 
-1. On GitHub, click the 'use this template' button to create a new repo on your account
-1. Clone the new repo to your computer
-1. Install required dependencies with npm: `npm install`
-1. Start the project: `npm start`
+The survey form is responsive and works in different browsers (Chrome, Firefox, Edge and Safari) and with ChromeVox Classic Extension.
```
sandra-shumanteva-kaevska, line 13: Great job. Even that wasn't requirement you did it.

code/src/Restart.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+
+const Restart = () => {
+  const resetSurvey = () => {
+    window.location.reload();
+  }
```
sandra-shumanteva-kaevska, line 7: You don't need to reload the page, by resetting the state variables into their initial values, the form will go to it's initial state. For example:
`setShowQuestion(false);
  setPlace('');
 setRating('');
  setFeedback('');
 setShowSummary(false);`

code/src/Submit.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const Submit = ({ onClick }) => {
```
sandra-shumanteva-kaevska, line 3: This Submit component is the same like Start component. You can combine both of them into one button component, that will also have a title like parameter

sandra-shumanteva-kaevska (commented):
You have very god job done. I think you understand components, props and state. But I also think that form logic should be simplified and some components can be reused, like button component.

code/src/Submit.js
MindstormingAB: Thank you so much for this comment Sandra! I replaced my 3 buttons with one unique button 🥳

code/src/Form.js
```diff
@@ -0,0 +1,80 @@
...
+      ) : (!showSummary && !showQuestion) ? (
+        <form>
+          <Intro />
+          <Start onClick={handleStart} />
+        </form>
+      ) : (
+                  <Summary place={place} rating={rating} feedback={feedback} />
+                )}
+    </section>
+  )
```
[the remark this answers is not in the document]
MindstormingAB: OMG! Great! I struggled a while with these conditions and your proposition is brilliant! 🙏

HIPPIEKICK (changes requested):
Changing to CR because of inconsistent use of semicolons, you should either have them everywhere or not at all :)

MindstormingAB:
@HIPPIEKICK Added semicolons everywhere (I hope 🤔)

HIPPIEKICK (approved):

PR #119, opened 2020-10-25
Week 10: Survey

monkeybusiness101:
Link to my deployed project:
https://awesome-jang-c9f2cb.netlify.app

Fun project. Nice to work with states and props more. 


code/src/Form.js
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react'
+import { Dropdown } from 'Dropdown'
+import { Radio } from 'Radio'
+import { Text } from 'Text'
+import { Submit } from 'Submit'
+
+
+
```
sofiavazs, line 8: same as above and no need for so many empty lines 🙈

code/src/Form.js
```diff
@@ -0,0 +1,72 @@
...
+export const Form = () => {
+ 
+    const [dropInput, fetchDropInput] = useState("")
+    const [radioInput, fetchRadioInput] = useState("")
+    const [textInput, fetchTextInput] = useState("")
+    const [section, setSection] = useState('start')
+
```
sofiavazs, line 15: nice for having all the "useState" in one place! 💃

code/src/Form.js
```diff
@@ -0,0 +1,72 @@
...
+                </section>
+            )}
+
+            {section === 'summary' && (
+                <section class="summary" tabIndex="0">
+                    <p tabIndex="0">Favo{radioInput}, take this boogie off my tail!</p>
+                    <p tabIndex="0">"I feel the need for {textInput}!"</p>
+                    <p tabIndex="0">You rated Top Gun:{dropInput}</p>
+                </section>
+            )}
```
sofiavazs, line 67: Nice multistep!  👏🏻

code/src/Radio.js
```diff
@@ -0,0 +1,15 @@
...
+export const Radio = ({ handleRadioInput }) => {
+  return <div className="question" tabIndex="0">
+      <fieldset>
+        <legend tabIndex="0">Which character has the best callsign?</legend>
+        <label for="mav" tabIndex="0">Maverick</label> 
+        <input type="radio" value="Maverick" id="mav" onChange={e => handleRadioInput(e.target.value)} required/>
+        <label for="goose" tabIndex="0">Goose</label> 
+        <input type="radio" value="Goose" id="goose" onChange={e => handleRadioInput(e.target.value)}/>
+        <label for="sundown" tabIndex="0">Sundown</label> 
+        <input type="radio" value="Sundown" id="sundown" onChange={e => handleRadioInput(e.target.value)}/>
```
sofiavazs, line 12: I can see that you added "required" in one the radio buttons, however I could go through the form without choosing any of them, I had the same issue and I solved it but also adding a "name"  and "checked" attributes 😊

sofiavazs (approved):
Great work Max! Congrats! 
Top gun is great! I gave it 10/10 in your survey 🤓 
Really nice that you decided to make it multistep! 👏🏻 Your code is clean and easy to read! 

I made some comments along the code and hope it helps, but some suggestions: 

- Since you made it multi step, with accessibility in mind would have been nice to have a welcome page explaining how to navigate the survey, so users that use screen readers would know how to proceed. 
- Still in accessibility, some "aria-labels would have made a slight difference on how users experience the survey. I also noticed that as I was tabbing along the survey the focus from one step to the next was always on the "next" button, one thing that you could have added was the `autoFocus` in the input, more info [here ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus)

These are just some suggestions that might help in the future, I think you did a great job!!

code/src/Dropdown.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+export const Dropdown = ( {handleDropInput} ) => {
+    return <div className="question">
+      <label for="Movie" tabIndex="0">Rate Top Gun</label>
+        <select name="Movie" id="Movie" onChange={e => handleDropInput(e.target.value)} tabIndex="0" required>
```
EmOlsz, line 6: All of those `Movie` values should be written lowercase

code/src/Form.js
```diff
@@ -0,0 +1,72 @@
...
+            {section === 'start' && (
+                <section>
+                    <p className="question">Stay on my wing and ill take you home?</p>
+                    <Submit title="Yes!" setSection={setSection} value='callsign'/>
+                </section>
+            )}
+
+            {section === 'callsign' && (
+                <section>
+                    <Radio handleRadioInput={handleRadioInput} />
```
EmOlsz, line 42: You are going to use that method inside `<Radio />` component, so it should be named `onRadioInput`, like this :
```
<Radio onRadioInput={handleRadioInput} />
```

code/src/Radio.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export const Radio = ({ handleRadioInput }) => {
+  return <div className="question" tabIndex="0">
+      <fieldset>
+        <legend tabIndex="0">Which character has the best callsign?</legend>
+        <label for="mav" tabIndex="0">Maverick</label> 
+        <input type="radio" value="Maverick" id="mav" onChange={e => handleRadioInput(e.target.value)} required/>
```
EmOlsz, line 8: Very nice idea in general, but you forgot to add `checked` attribute. Because of that, I was able to check all 3 radio buttons (and I should select only one at the time)

EmOlsz:
Hey Max, good job for your project! Take a look at my comments and provide small changes

code/src/App.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react'
+import { Form } from 'Form'
+import { Header } from 'Header'
```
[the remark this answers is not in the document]
monkeybusiness101: Fixed, thank you!

code/src/Form.js
monkeybusiness101: Best movie!

code/src/Dropdown.js
monkeybusiness101: fixed

code/src/Form.js
monkeybusiness101: I did not fully understand this comment but I changed all props to onRadioInput, onTextInput, onDropinput. I thought it was ok to name the prop to be the same as the parameter?

code/src/Form.js
monkeybusiness101: fixed

code/src/Radio.js
monkeybusiness101: Fixed. I forgot to push where the radio buttons where in the same group or fieldset

code/src/Radio.js
monkeybusiness101: I saw your solution and that was really nice! I do not know exactly what I am doing wrong but I can not get the required to work

monkeybusiness101:
@sofiavazs Thank you for the feedback! I made some adjustments. I tried to get the required to work but to no avail
@EmOlsz I have made appropriate changes. Thank you for the feedback!

PR #126, opened 2020-10-25
Pull request for React survey project

Katsiez:
Link to my deployed project: https://self-care-journal.netlify.app/

My reflections on how this week's project turned out: It was a tough one and despite having validated the form, the user is still able to skip through questions and I would essentially like help to fix that. 

Things I'd like to have clarified or explained in more detail: Form validation, more specifically, so that user cannot skip through questions.


code/src/components/SecondQuestion.js
```diff
@@ -0,0 +1,28 @@
...
+                    id="answer"
+                    type="text"
+                    value={answer}
+                    onChange={event => setAnswer(event.target.value)}
+                    minLength="10"
+                    placeholder="Be kind to yourself"
+                    required
+                />
+            </label>
+            <button><a href="#hours">continue</a></button>
```
jonnaru, line 24: I think you have problems with required because you have these buttons without onSubmit. Maybe you know that :) Really tricky because you don't want to submit before the last submit-button. I have no idea, this is just a guess, but maybe try something like changing the button-tags to divs/spans or something? 🤷‍♀️

code/src/components/outputSummary.css
```diff
@@ -0,0 +1,41 @@
...
+  background-color: #efd5c9;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  padding: 20px;
+  margin: 0;
+}
+button.submit {
+  padding: 0 0;
```
jonnaru, line 11: Oops, a double zero!

code/src/components/secondQuestion.css
```diff
@@ -0,0 +1,29 @@
...
+  justify-content: center;
+  padding-top: 20px;
+}
+input[type="text" i] {
+  padding: 5px 0;
+  width: 500px;
+  height: 30px;
+  font-size: 18px;
+  font-style: italic;
+  border-radius: 0;
```
jonnaru, line 16: Do you need this?

jonnaru (approved):
Great work, Katarina 🤩 The code is structured, clean and easy to follow. Super cool. I really like how you styled it with the different scroll-pages and the colours. Cute radio buttons:) I hadn't much to say, just some small picky things. Hope you'll find a solution with the requirements!

code/src/components/FourthQuestion.js
```diff
@@ -0,0 +1,41 @@
...
+                        required>
+                        <option value="" disabled>♡</option>
+                        <option value="try out a nice and relaxing meditatation">Meditate</option>
+                        <option value="drink more water, have a glass right now">Drink more water</option>
+                        <option value="go on a long walk, maybe go with a loved one">Go on a long walk</option>
+                        <option value="journal your thoughts and feelings on a regular basis">Journal your thoughts and feelings</option>
+                        <option value="enjoy a cup of tea">Enjoy a cup of tea</option>
+                        <option value="take that warm and cozy bath">Take a warm bath</option>
+                    </select>
+                    <button type="submit"
```
HIPPIEKICK, line 32: One option might be to conditionally render/disable the buttons. So if the activity is not set, that's a false by default and you can use that either by:
`{activity && <button>}`
or
`<button disabled={!activity}`

HIPPIEKICK: Same for the other components - forcing a state change before continuing

PR #133, opened 2020-10-25
Craft Beer Co. survey by Claire Caudwell

ClaireCaudwell:
Link to my deployed project: https://flamboyant-poitras-c5efb2.netlify.app/

My reflections on how this week's project turned out: Please see my [README](https://github.com/ClaireCaudwell/project-survey)

Things I'd like to have clarified or explained in more detail: Please see [README](https://github.com/ClaireCaudwell/project-survey)


VanessaSue27 (changes requested):
Hello Claire! 😁
I am so inspired after doing your code review!
I can tell how much work, love and time you have put into this project.
There are several small pieces of logic implemente here and there, just to make sure things are displayed properly: like the "is" or "are" and the adding of a "," or "&". Those take extra time to add, but having these things implemented make a big difference when navigating the survey.
You have inspired me to be more thorough as well 💪
I don't have much feedback about the use of State, I think your code was clean and easy to follow, the comments made every step clear, hwo to start and move on.
I can also notice that you are very design-oriented and a lot of time went into the CSS 😍 A lot of care into customizing inputs instead of leaving the default styles, so that it all falls into place with your theme and colors, super well done!

My not super positive feedback is mainly about the Accessibility, which you already mentioned you still need to work on. Besides the comments I left, here are some other things I noticed:
- User can move on to the next question without having entered an answer on the current question.
- When doing the keyboard testing: I noticed mainly problems in the radio buttons and checkboxes: I can't move through them using the arrow keys.
- When doing the screen reader testing with ChromeVox: in checkboxes, the labels are being read twice.

I will have to ask for Changes to be done, mainly from the accessibility point of view, since there are some glitches.
But in my opinion, from the React and use of State angle, you've done an amazing job! 🎉

/Vanessa 🌴

code/src/App.js
```diff
@@ -1,9 +1,50 @@
...
+      <main>
+        { pageState === true ? (
+          <Welcome setPageState={setPageState}  />
+        ) : (
+          <Form />
+        )}
+      </main>
+      {/* <footer className="footer-container">
+        <p>Webpage created by Claire Caudwell using React and React state hooks. Part of the Technigo Bootcamp 2020</p>
+      </footer> */}
```
EmOlsz, line 45: These 3 lines of code could be deleted before final push

code/src/Form components/Name.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+export const Name = ({name, setName}) => {
+
+    return (
+        <fieldset className="container">
+            <legend tabIndex="0">Please enter your name</legend>
+            <label for="name">
```
EmOlsz, line 8: In React files we need to use `htmlFor` instead of normal `for` attribute - it's because `for` is a reserved keyword in JavaScript and we need to remember that JSX is actually normal JS, not HTML ;)

code/src/Form.js
```diff
@@ -0,0 +1,69 @@
+import React, {useState} from 'react'
+import { Name } from './Form components/Name.js'
```
EmOlsz, line 2: Next time let's remember about empty space between outer and local dependencies

code/src/Summary.js
```diff
@@ -0,0 +1,61 @@
...
+            if (array.length-1 === i) {
+                
+            } else if (array.length-2 === i) {
+                string += " & ";
+            } else {
+                string += ", "
+            }  
+        }
+        return string;
+    };
```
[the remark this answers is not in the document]
EmOlsz: Indeed, great function :)

ClaireCaudwell:
@VanessaSue27 @EmOlsz Thanks both for the reviews! I worked alot on this project, but it is a WIP to tidy things up and get the accessibility working. It will be my side project whilst working on the weekly projects as I want to make it better! Thank you both for all your feedback, advice and tips. I will use them when and make the changes accordingly.

PR #138, opened 2020-10-26
Rebeka: Project-survey

rebkova:
Link to my deployed project: https://rebeka-project-survey.netlify.app/

I'm happy with what I managed to do under such short time I had available this week. I had struggled with second deploy to Netlify quite a bit ... Fixed it at the end so all is good.


code/src/components/Form.js
```diff
@@ -0,0 +1,94 @@
...
+import Checkbox from "components/Checkbox";
+import Text from "components/Text";
+import DropDown from "components/DropDown";
+import RadioButton from "components/RadioButton";
+import Submit from "components/Submit";
+import Summary from "components/Summary";
+import Reload from "components/Reload";
+
+import "./form.css";
+
```
Corneliaryas, line 13: Great structure with the line breaks between imports!

code/src/components/Form.js
```diff
@@ -0,0 +1,94 @@
...
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    setDisplayForm(false);
+    setDisplaySummary(true);
+  };
+
+  return (
+    <form className="app-container" onSubmit={handleSubmit}>
+      {displayForm && (
+        <div className="form-container">
```
Corneliaryas, line 51: Interesting to have the form tag outside of the conditional and wrapping both questions and the summary, with a div tag that sets the design of the form. Does the summary need to be inside the form tag? Otherwise maybe wrap the whole thing in a section and just the form itself in a form tag, just to make it a little bit easier to follow along. 😊

code/src/components/Header.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+import "./form.css";
+
+const Header = ({ title, subtitle }) => {
+  return (
+    <section className="section-container">
+      <h1 tabIndex="0">{title}</h1>
```
Corneliaryas, line 8: Do you need the tabIndex here? I think the H1 is enough for accessibility unless you can interact with the title

code/src/components/RadioButton.js
```diff
@@ -0,0 +1,35 @@
...
+
+const RadioButton = ({ radioText, radioAnswer, onAnswerChange }) => {
+
+  return (
+    <section className="section-container">
+      <h2>{radioText}</h2>
+      <div className="radio-btn-container">
+        {answersArray.map(item => {
+          return (
+            <label tabIndex="0" key={item} htmlFor={item}>
```
Corneliaryas, line 16: Same here with the htmlFor, I think you can skip it sense the input field is inside the label tag?

code/src/components/Reload.js
```diff
@@ -0,0 +1,25 @@
...
+
+  const reloadPage = () => {
+    window.location.reload();
+  };
+
+  return (
+    <section className="section-container">
+      <button
+        tabIndex="0"
+        type="reset"
```
Corneliaryas, line 15: Nice, I didn't know about the reset type! 🙌

code/src/components/Summary.js
```diff
@@ -0,0 +1,18 @@
...
+
+const Summary = ({ isBoxChecked, textAnswer, dropdownAnswer, radioAnswer }) => {
+  return (
+    <div className="section-container">
+      <h1 tabIndex="0">Summary of your moving requirements:</h1>
+      <p>Thank you for your interest, {textAnswer}!</p>
+      <p> &#9745; You want price for {isBoxChecked}.</p>
+      <p>&#9745; You're moving to a new place which is {dropdownAnswer} away.</p>
+      <p>&#9745; You said {radioAnswer} to renting cardboard boxes.</p>
+      <p>A price estimation will be sent to you shortly.</p>
```
Corneliaryas, line 13: Maybe consider naming the states to a more describing name, so it would make more sense here for example, if you are not that involved with the code :)  (for example textAnswer could be "name" and dropDownAnswer could be "kilometers" or something) 😊

code/src/components/Text.js
```diff
@@ -0,0 +1,26 @@
...
+      <label tabIndex="0" htmlFor="textField"><h2>{questionText}</h2></label>
+      <input
+        tabIndex="0"
+        type="text"
+        id="textField"
+        name="answerOne"
+        autoComplete="off"
+        required
+        //the bottom 2 parts & the type are controlled components
+        //arrow functions are less efficient, better to use function name only
```
Corneliaryas, line 19: consider removing comments ;)

code/src/components/dropdown.css
```diff
@@ -0,0 +1,7 @@
+.form-container {
+  display: flex;
+}
```
Corneliaryas, line 3: this is also set in the form.css, is it needed in both files?

code/src/components/form.css
```diff
@@ -0,0 +1,68 @@
...
+.section-container {
+  display: flex;
+  flex-direction: column;
+  padding: 40px 270px;
+  border-bottom-style: dashed;
+}
+
+.section-container:last-child {
+  border-bottom-style: none;
+}
```
Corneliaryas, line 27: Nice!

Corneliaryas (approved):
Super good job Rebeka! Really like the site and the theme! 😍😊 Great clean and nice-looking code!
Sorry, I couldn't answer the questions you had in the comments, hopefully, a teacher can!

code/src/components/Checkbox.js
```diff
@@ -0,0 +1,49 @@
...
+      <h2 tabIndex="0">I would like a price estimation for:</h2>
+      <label htmlFor="moving">
+        <input
+          tabIndex="0"
+          type="checkbox"
+          id="moving"
+          onChange={() => onCheckBoxChange("moving")}
+          checked={isBoxChecked.includes("moving")}
+        >
+        </input>
```
EmOlsz, line 19: `<input />` is self closing tag, no need to have an opening and closing tag separate

code/src/components/Checkbox.js
```diff
@@ -0,0 +1,49 @@
...
+        Packing
+      </label>
+      <label htmlFor="cleaning">
+        <input
+          tabIndex="0"
+          type="checkbox"
+          id="cleaning"
+          // onChange={(event) => onCheckBoxChange(event.target.checked)} - what is the difference?
```
EmOlsz, line 38: This `onChange` event will update `checkBox` state property from `<Form />` component with value `true` or `false` - depends on if you uncheck or check this input.
Line below, `onChange={() => onCheckBoxChange("cleaning")}` will update state property with string value `cleaning`. I assume you want the second line to be executed rather than the commented one :)

code/src/components/DropDown.js
```diff
@@ -0,0 +1,36 @@
...
+            value={dropdownAnswer}
+            required
+            id="drop-down"
+          >
+            <option value="">-</option>
+            <option tabIndex="0" value="0-10 km">0-10 km</option>
+            <option tabIndex="0" value="20-50 km">20-50 km</option>
+            <option tabIndex="0" value="50+ km">50+ km</option>
+
+            {/* why "required" attribute doesn't work with <select> when I use map()? */}
```
EmOlsz, line 25: I think it does, the issue might be connected to the fact that you initialise `distance` state property in `<Form />` component as empty object `{}`, but later on if you select any value, it will be string

code/src/components/Form.js
```diff
@@ -0,0 +1,94 @@
...
+  const [checkBox, setCheckBox] = useState([]);
+  const [text, setText] = useState("");
+  const [distance, setDistance] = useState({});
+  const [answer, setAnswer] = useState("");
+
+  //all state functions
+  const handleCheckBoxChange = (newCheck) => {
+    checkBox.includes(newCheck)
+      ? setCheckBox(checkBox.filter(item => item !== newCheck))
+      : setCheckBox([...checkBox, newCheck]);
```
EmOlsz, line 27: Great job :)

code/src/components/Text.js
EmOlsz: This is very descriptive comment, should definitely stay

PR #146, opened 2021-03-18
Ylva Landoff Lindberg Survey

sssylvasss:
Link to my deployed project:
https://face-survey-ylva-landoff-lindberg.netlify.app

It went ok. I think I understod what I did. Got help from friends and 1 on1 .


code/src/App.js
```diff
@@ -1,9 +1,42 @@
...
+
 
 export const App = () => {
+const [name, setName] =useState();
+
+const [sizeOfNose, setSizeOfNose] = useState("")
+
+const [eyeColor, setEyeColor] = useState("");
+
+const [smile, setSmile] = useState()
```
malinvan, line 19: Should this also have a default empty string as the other face-states? I suppose they should all have the same default value though.

code/src/App.js
```diff
@@ -1,9 +1,42 @@
...
+
+const [hasMustache, setHasMustache] = useState("");
+
+const [submit, setSubmit] = useState();
+
   return (
-    <div>
-      Find me in src/app.js!
```
malinvan, line 6: Remember to remove the code from the boiler plate :)

code/src/App.js
```diff
@@ -1,9 +1,42 @@
...
+    <div className="main-container">
+      {submit ? (
+        <Respons name={name} sizeOfNose={sizeOfNose} eyeColor={eyeColor} smile={smile} hasMustache={hasMustache}/>
+      ) : (
+      <form submit={submit}>
```
malinvan, line 30: I love how clean your code looks, it's easy to follow.

code/src/Components/NoseQuestion.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+
+
+export const NoseQuestion = ({ sizeOfNose, setSizeOfNose }) => {
```
malinvan, line 4: Here I think you could have used an iteration, props and an array in the App.js to only have one block of code in this component rather than hardcoding the three radios buttons.

code/src/Components/SmileQuestion.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+
```
malinvan, line 2: Same here as with previous radio buttons

code/src/index.css
```diff
@@ -1,13 +1,186 @@
+
+
```
malinvan, line 2: Just remember to clean the CSS up a bit for readability. Remove empty spaces in some paces and add a break in other so it's easier to read.

malinvan (changes requested):
Again I really like your creativity and your clean code. The only thing I can see to be changed is the accessibility which is a requirement for the blue level. I tried it with the voice over and tabbing and it didn't work everywhere on the site.

But again I love it and can't wait to see your future projects :)

code/src/App.js
```diff
@@ -1,9 +1,42 @@
-import React from 'react'
+import React, { useState }  from 'react'
+import { NameQuestion }  from './Components/NameQuestion.js';
+import { NoseQuestion }  from './Components/NoseQuestion.js';
+import { EyeQuestion }  from './Components/EyeQuestion.js';
+import { SmileQuestion }  from './Components/SmileQuestion.js';
+import { MustacheQuestion }  from './Components/MustacheQuestion.js';
+import { Heading }  from './Components/Heading.js';
+import { SubmitQuestions }  from './Components/SubmitQuestions.js';
+import { Respons }  from './Components/Respons.js';
```
KarolCislowski, line 9: Remember to separate imports Components that are from node_modules, and your's with one empty line.

code/src/App.js
KarolCislowski: it would be nice, but not necessary here 🦄

KarolCislowski (approved):
Great job! 

PR #150, opened 2021-03-19
Week 10 - Project Survey

julialeonieros:
Link to my deployed project:
https://focused-sammet-d2a2f0.netlify.app/

I think it turned out pretty good, considering that I got really stuck with showing and hiding form and summary. Solved it eventually with some help from a teammate. 

I first tried to use html-validation on the form ('required'), but somehow that didn't work, even though it's all in the same form and I have a submit-button. So instead I did a function to test if the inputs are empty at submit. 

I deployed the page before I was finished with the styling, hope it will be updated after the pull request is approved (?)


code/src/App.js
```diff
@@ -1,9 +1,76 @@
...
+      return false
+    } else {
+      return true
+    }
+  }
+
+  // refreshes page when button on summary-page is clicked
+  const refreshPage = () => {
+    window.location.reload()
+  }
```
Pauan8, line 35: Neat solution to clear all inputs 👍

code/src/components/Button.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export const Button = (props) => {
+  return (
+    <div className='submit-button-container'>
+      <button onClick={props.onClick} className='submit-button' type='submit' form=''>{props.title}</button>
+    </div>
```
Pauan8, line 7: Instead of using {props.title} we should only write {title} and do:
`export const Button = ({title}) => {`

or like this:

`export const Button = props => {`
and:
`const {title} = props`

code/src/components/Form.js
```diff
@@ -0,0 +1,82 @@
...
+            id="genre"
+            value={genre}
+            onChange={(e) => {setGenre(e.target.value)}}
+          >
+            <option disabled></option>
+            <option value="biography">Biography</option>
+            <option value="classics">Classics</option>
+            <option value="fiction">Fiction</option>
+            <option value="comedy">Comedy</option>
+            <option value="crime">Crime</option>
```
Pauan8, line 27: When writing repeated code like this - an array and array.map can also be used!

code/src/components/Form.js
```diff
@@ -0,0 +1,82 @@
...
+          <label htmlFor="language">Select preferred language:</label>
+          {languages.map(lang => (
+            <label key={lang}>
+              <input 
+                id="language"
+                type="radio"
+                value={lang}
+                checked={language === lang}
+                onChange={(e) => {setLanguage(e.target.value)}}
+              />
```
Pauan8, line 52: Here the id for each input is "language". We should not use the same id for different elements.
Instead htmlFor can be added to the label inside the map.
Then id + htmlFor can be set to {lang}.

code/src/components/Form.js
```diff
@@ -0,0 +1,82 @@
...
+
+const languages = [' english', ' swedish', ' both'] // array for radio-buttons under language
+
+export const Form = (props) => {
+  const { genre, setGenre } = props
+  const { favorite, setFavorite } = props
+  const { language, setLanguage } = props
+  const { name, setName } = props
+  const { email, setEmail } = props
+
```
Pauan8, line 11: It's not needed to have these divided. 
We can write: 
` const { genre, setGenre, `
`favorite, setFavorite,`
` language, setLanguage,  `
`name, setName, `
`email, setEmail } = props`

code/src/components/Form.js
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <div className="form-container">
+      <form>
+        <div className="form-container-sections">
```
Pauan8, line 15: Good CSS BEM-like naming 🥇

Pauan8 (approved):
A great project that works 100% as intended 💯 The code is clean and easy to follow.

I only have some minor comments for you to look at for future projects!

KarolCislowski (approved):
Great job! 
For the next step, it would be nice to try separate questions to multiple pages. 🌈

PR #154, opened 2021-03-19
project-survey-Ati Takrami

takrami:
https://blissful-ardinghelli-ea12d4.netlify.app

I really enjoyed working with react. I struggled a lot with structure of project and the components' name.


code/src/assets/styles/global.css
```diff
@@ -0,0 +1,101 @@
...
+  left: 0;
+  top: 0;
+  width: 18px;
+  height: 18px;
+  border: 1px solid #ddd;
+  border-radius: 100%;
+  background: #fff;
+}
+[type="radio"]:checked + label:after,
+[type="radio"]:not(:checked) + label:after {
```
Sartish, line 76: Nice! I didn't know you can style it this way before and after the box is checked🙋‍♀️

code/src/assets/styles/global.css
```diff
@@ -0,0 +1,101 @@
...
+  top: 4px;
+  left: 4px;
+  border-radius: 100%;
+  -webkit-transition: all 0.2s ease;
+  transition: all 0.2s ease;
+}
+[type="radio"]:not(:checked) + label:after {
+  opacity: 0;
+  -webkit-transform: scale(0);
+  transform: scale(0);
```
Sartish, line 91: Maybe an idea, to add some small comments pseudo code to know what the code does. Not everywhere of course,  but maybe where there are some animations etc. Maybe it will be hard to remember everything later when looking back at project, what was the end goal of the animation etc. Just an idea 😄

code/src/assets/styles/shared.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+  &:hover ${ButtonText} {
+    color: #fff;
+  }
+`;
+
+const ButtonsContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  margin-top: calc(var(--spacing) * 16);
```
Sartish, line 27: ahh love your use of grid <3 Not sure what this calc(var--spacing) does. But it looks so neat!  Will google this later

code/src/components/experience-type/experience-type.style.js
```diff
@@ -0,0 +1,11 @@
+import styled from "styled-components";
+
+const ClimbingTypeSelect = styled.select`
```
Sartish, line 3: Is this a typescript thing? Or a CSS thing? Looks like a good way of using variables and then style them in a way way👏

code/src/App.js
```diff
@@ -1,9 +1,24 @@
...
+import Survey from "./components/survey/survey.component";
+import Main from "./pages/main/main.component";
+import NotFound from "./pages/not-found/not-found.component";
+import Footer from "./components/footer/footer.component";
+
+function App() {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
```
Sartish, line 7: Maybe take this away part away ☺️?

code/src/components/climbing-area/climbing-area.component.jsx
```diff
@@ -0,0 +1,43 @@
...
+    <>
+      <Label>What type of area for climbing do you prefer?</Label>
+      {areas.map((climbingArea) => {
+        return (
+          <AreaOption
+            area={climbingArea}
+            setArea={setArea}
+            key={climbingArea} />
+        );
+      })}
```
Sartish, line 30: Nice! I really like this functions! both here and live in the site. Well done 🙋‍♀️

code/src/components/experience-duration/experience-duration.component.jsx
```diff
@@ -0,0 +1,43 @@
...
+      <Progress value={range} max="100" id="progress" />
+      <RangeInput type="range" min="0" max="100" onChange={onRangeChange} />
+      <ButtonsContainer>
+        <ButtonContainer type="submit">
+          <ButtonText to="/step-3">Previous</ButtonText>
+        </ButtonContainer>
+        <ButtonContainer type="submit" onClick={onNextHandler}>
+          <ButtonText to="/step-5">Next</ButtonText>
+        </ButtonContainer>
+      </ButtonsContainer>
```
Sartish, line 38: SOOOO COOOOL!🤯😎

code/src/components/result/result.component.jsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return (
+    <>
+      <div style={{ backgroundColor: `${backgroundColor}`, padding: "16px", display: "grid", gap: "16px", marginBottom: "32px" }}>
+        <p style={{ color: `${textColor}` }}>{userName}, you chose {climbingType}. Luckily you have {equipments}. Also you {experienceText()} You prefer {area} climbing.</p>
+      </div>
+      <ButtonContainer onClick={onResetForm}>
+        <ButtonText to="/step-1">Start Again</ButtonText>
+      </ButtonContainer>
```
Sartish, line 50: I tried this start again in your deployed  netfliy app, but it didn't start from the beginning. Maybe it is something with netlify? Otherwise very good idea. And well done on localstorage⭐️. I find this very difficult to use myself.

Sartish (approved):
Hi Ati! 
As always I am so impressed by your code⭐️. You are very structured in the way you are coding. Which is according to me the most difficult things in coding! You are always trying to use code effectively and find the shortest  and most effective way of writing the code. Which is so great🤩
I put the project as approved and comments are only there as suggestions or "wow" comments (mainly in css 💌) Again amazing work, you are awesome 🥳

code/src/assets/styles/global.css
takrami: Thanks Sara! definitely! I'll do it :+1:

code/src/assets/styles/shared.js
takrami: The calc() CSS function lets you perform calculations when specifying CSS property values. In this project I defined a variable as spacing then used it with calc function in different places. It can help us to have more control on design. For example if we want to increase or decrease the spaces in whole project it needs just be changed in one place (where we define spacing variable).
You can read more about calc here: https://developer.mozilla.org/en-US/docs/Web/CSS/calc()

code/src/components/experience-type/experience-type.style.js
takrami: This is styled-components actually a CSS thing.  Here you can read about some reasons for use it and also how to use it. https://styled-components.com/docs/basics#motivation

code/src/App.js
takrami: Oh! I forgot it! Thanks to mention it. :pray:

code/src/components/result/result.component.jsx
takrami: I'll check it again. Thanks :+1:  :pray:

KarolCislowski (approved):
Great job! 

PR #156, opened 2021-03-20
React-Survey-Project-10-Shaun-Newmark

shauninc:
Link to my deployed project:
https://react-basic-survey-sn.netlify.app/

Was great to practice with state and different approaches to solve how to render each component

Conditional rendering and state, is there a prefered  approach to render only one component at a time ? 

Is the convention to add this conditional rendering function within the app.js file ?


code/src/components/Question1.js
```diff
@@ -0,0 +1,26 @@
...
+    <>
+    <p>Start by entering your name  
+        <span role="img" aria-label="Smile Face Emoji">👇</span>  
+    </p>
+            <div>
+                <label htmlFor="name">Name: </label>
+                <input
+                    type='text'
+                    onChange={(e) => onNameChange(e)}
+                    required
```
katliebig, line 15: The basic HTML validation is working fine for both the text input and the select!

code/src/components/Question3.js
```diff
@@ -0,0 +1,26 @@
...
+        <>       
+        <p>Preferred Investment:</p>
+                 { investmentTypes.map(item => (
+                     <label key={item}>
+                        <input
+                            type="radio"
+                            name="investmentType"
+                            value={item}
+                            onChange={(e) => onInvestmentTypeChange(e)}
+                        />
```
katliebig, line 16: You could add the required-attribute here as well. Right now, the user doesn't have to choose a preferred investment.

code/src/components/Summary.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+const Summary = ({ name, checked, value }) => {
+
+    const restart = () => {
```
katliebig, line 5: Really smart solution to restart!

code/src/components/Question1.js
```diff
@@ -0,0 +1,26 @@
...
+const Question1 = ( {name, onNameChange}) => {
+
+    return (
+    <>
+    <p>Start by entering your name  
+        <span role="img" aria-label="Smile Face Emoji">👇</span>  
+    </p>
+            <div>
+                <label htmlFor="name">Name: </label>
+                <input
```
katliebig, line 12: You forgot to add the id to the input to match it with the label.

code/src/index.css
```diff
@@ -1,13 +1,84 @@
+@import url("https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@100&display=swap");
```
katliebig, line 1: The styling is nice, I like the emojis. One suggestion is to put some padding to the inputs to increase accessibility. They can get quite small, especially on mobile.

code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+          />
+       
+        <button className="submit" type="Submit">Submit</button>
+
+        </form>
+    )}
+
+      {showSummary && (
+        <Summary name={name} checked={investmentType} value={investmentTerm} />
+      )}
```
katliebig, line 83: One solution to render one question at a time could be this: 

First, you'll need to add a hook for the section, starting at 1:
`const [section, setSection] = useState(1);`

This state will be changed by clicking on a button:
`const onButtonClick = () => {
    setSection(section + 1);
  };`

And this button will be rendered with each question, except for the last one, where your submit button will be rendered.

This is what will be rendered (minus the header). I've excluded the props for the questions and the summary components, don't forget them ;)

```
if (!showSummary) {
    return (
      <div>
        {section === 1 && (
          <div>
            <Question1 />
            <button onClick={onButtonClick}>Next Question</button>
          </div>
        )}

        {section === 2 && (
          <div>
            <Question2 />
            <button onClick={onButtonClick}>Next Question</button>
          </div>
        )}

        {section === 3 && (
          <div>
            <Question3 />
            <button onClick={handleSubmit}>Submit</button>
          </div>
        )}
      </div>
    );
  } else if (showSummary) {
    return <Summary />;
  }
```

This solution builds on what you've already used with the showSummary hook. But there is one downside to it, your validation won't work anymore - it's working right now because it's basic HTML validation within a form, but now your question won't be in a form anymore. You could solve that by adding custom validation (have a look at Van's lecture again), by writing some extra logic that's executed when the next question button is clicked - basically checking if the value of the current input is empty or not.

katliebig (changes requested):
Really nice project! I like the emojis 👍

There's one small thing you missed: the label for the name input is not connected, because the input is missing the matching id. As it's one of the requirements to follow accessibility guidelines, you get changes requested. No big deal though! :)

Besides from that, I wrote a suggestion on how to render only one question at a time. It might be over the top, and your project is working just fine, but you can try it out if you like.

petercreations:
@katliebig Thank you for the feedback and suggestions, super helpful ! 🙂  Added the changes  👍

KarolCislowski (approved):
Great job!

For the next step, it would be nice to try separate questions to multiple pages.

PR #158, opened 2021-03-20
Project-survey-Estefanía

esteficodes:
Link to my deployed project:

https://project-survey-estefania.netlify.app/

This week's project started out very well. I decided to go step by step, trying to understand one thing at a time. To me it was really clarifying to see the image of the React flow that we should follow because that helped me to understand that I needed to handle everything from the app component, as if this was "the brain" of the project. There was when I redirected my approach and was able to correct misplancing of elements, resulting in a good functioning app.

I was also maybe too ambitious when trying to solve the progression of the sections, but I ended up making it work with a back and next button in each section.

I would like to have the mapping explained more in detail. This took me a long time of trial and error (and help from my team) to figure out. The concepts of "key" and "id" are not enterely clear to me. I would also need more instruction into how to style forms because this is an area that I haven't had time to focus on yet.


code/src/components/MainDish.js
```diff
@@ -0,0 +1,25 @@
...
+            <select 
+                className='maindish-menu'
+                name={label}
+                value={mainDish}
+                onChange={onMainDishChange}
+                >
+                 <option>pasta</option>
+                 <option>meat</option>
+                 <option>seitan</option>
+                 <option>chicken</option>
```
TinaJuren, line 19: When actively choosing a dish the users choices are nicely printed in the summary 😃, but when the user doesn't make an active decision and leave the default choice (pasta) the state doesn't update and "pasta" is not printed in the summary. 
I checked my code and I have added something like: 
<option value=' '>Welcome to choose your dish</option> above my "pasta" option. I'm not sure if my solution is best practice though. However, Van provides another solution, using "disabled" for the first <option>, with no text. He describes it in Week10, lecture#2 about 19 minutes in 😃.

code/src/components/SideDish.js
```diff
@@ -0,0 +1,26 @@
...
+            <select 
+            className='sidedish-menu'
+              name={label}
+              value={sideDish}
+              onChange={onSideDishChange}
+              >
+              <option>french fries</option>
+              <option>beans and onion salad</option>
+              <option>leafy greens</option>
+              <option>peas and carrots</option>
```
TinaJuren, line 20: Same as for Maindish - an active choice has to me made to make it show in the summary.

TinaJuren (approved):
Your Dinner plans generator is a pleasure to use when making an order 😃. I'm inspired by your code solutions and structure, and that you created many components and back/next buttons. 👍The first and last impression made me forget about the lack of styling of the questions. You did a great job Estefanía👏

KarolCislowski (approved):
Great job! 

PR #159, opened 2021-03-20
week 10 survey project

Yennie-Leek:
Link to my deployed project:

https://heuristic-albattani-176af8.netlify.app

My biggest problem is that I still struggling with JS and now diving in to React. I have a hard time to understand the connections. 

This week has been a little bit better as I have gotten some help and guidance from my code coach. 
I did the first part on my own, but then got some help to connect the components to the Form and App files to be able to send data properly. 
Its been a lot of googling, watching the videos from Technigo and most of the help I got from my code coach. 

If I had more time I would go through the accessibility part more and try to make it optimised. There is some things that worked earlier (radiobuttons) and after some other changes it stop working and I can find out what the problem is. So I need to put some extra time in to it later to fix it.
Also I would have wanted to experiment with more functions, like range, progressbar and similar.


code/src/App.css
```diff
@@ -0,0 +1,449 @@
...
+  background: url("/assets/icecream.jpg")center center;
+  opacity: 0.5;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0; 
+  position: fixed; 
+  z-index: -1;
+  background-size:cover;
+  height: 100vh;
```
annalenksjo, line 12: This is interesting! I guess this is a part of your solution to show the last summary? What is it ::after really? I have mostly seen this on links where the after is then after click for example. Would be nice to add an explanatory comment?

code/src/App.css
```diff
@@ -0,0 +1,449 @@
...
+}
+
+/* FORM */
+
+.form-container {
+height: 100%;
+display: flex;
+flex-direction: column;
+align-items: center;
+justify-content: space-around;
```
annalenksjo, line 26: the declarations should be indented :)

annalenksjo: You have mixed a bit with indented and not indented so just correct where its not

code/src/App.css
```diff
@@ -0,0 +1,449 @@
...
+  .name-box {
+    width: 60%;
+    margin: 20px 50px 20px 50px;
+  }
+
+  h2 {
+    font-size: 40px;
+  }
+
+  .fill-in-field {
```
annalenksjo, line 375: Maybe add the same structure with headings in the media queries as well? Makes it a lot easier to follow!

code/src/App.css
```diff
@@ -0,0 +1,449 @@
...
+  display: flex;
+}
+
+.again-btn {
+  margin: 10px;
+  font-family: 'Monaco';
+  background-color: #AD8841;
+  color: #37372C;
+  border: solid 1px #824947;
+  height: 30px;
```
annalenksjo, line 264: remember that you can set css declarations for multiple classes if they have the same styling(or in part the same styling) so you can save some space! For example:
.again-btn,
.ice-pic {
   border:
   }
and then make the individual declarations on respective. Now it was only one in common in this example hehe but there are some classes above with more declarations in common !

code/src/Form.js
```diff
@@ -0,0 +1,59 @@
...
+      return false
+    }
+
+    if (timeInterval === '') {
+      return false
+    }
+
+    if (favorite === '') {
+      return false
+    }
```
annalenksjo, line 25: I dont think you need this part since all the fields are required so its already checked at that point!

annalenksjo (changes requested):
Hello Yennie! 
Really nice and colorful project, love that! I'm not the biggest ice cream lover my self though :P 
Everything seems to work as it should! 

Some things I suggest editing:

- Add cursor:pointer effect on the buttons
- Really like the background image! Although it make the text sligtly  difficult to read, especially if you have problems with vision   (especially the radio buttons). So maybe improve this in some way like   a background color on the text.
- Why is polka-option bold in the drop down? Is this intentional? I'm  curious why in that case - is it your favourite maybe? :P
 - Maybe remove the isResultCOmplete function I dont think it does  anything?



You mentioned that you had problems with your radiobuttons but I guess you fixed it because they worked fine for me! Maybe you will have time at some point to go back and add some of the extra functionality that you mentioned :) I wish the same for my project! Well done with 16 commits! You are really getting in to that habit! Well done!!

KarolCislowski (approved):
Great job!
Maybe for the next step, you will try to add some more exotic types of inputs? 🍹 🍨

PR #162, opened 2021-03-21
Finished the project Caroline Byström

Frivridna:
Link to my deployed project: https://postcovid19-survey-stage-art.netlify.app/

I would have liked to customize my Survey more, doing a header and a footer and perhaps a progress bar to show the user where in the survey they were. But as it is right now I am ok with the result. 

Mentioned in the README: 
"Wednesday's code coaching and lecture clarified a lot of things to me! :) But I found it especially hard to deal with the checkbox multiple choices. If I had more time I would style for tablet mode, and figure out how to limit user choices on checkboxes. This week I have a real lack of time, since I am working a lot at my breadjob so all of those things needs to be taken care of later."

How to work with checkboxes - can I only make it possible for a user to select a limited amount of checkboxes? 
Other input types that I did not try out this time.
How to properly work with arrays and objects in React.


code/src/components/QuestionStageArtCheckbox.js
```diff
@@ -0,0 +1,69 @@
...
+    "Concert",
+    "I just miss the feeling of collectivity, and long to be able to see something with another person's eye - just once, please!", 
+    "I want to drink beer with my friends after a show"
+  ]
+
+  return (
+    <article className="form-question-3">
+      {/* Q */}
+      <p className="form-question" tabIndex="0">
+        What kind of stage art would you like too experience post Covid-19? Pick as many as you like! <span role="img" aria-label="smiling emoji with one eye blinking">😊</span>
```
liaback-lowstett, line 35: Emojis is always a good idea 😄

code/src/index.css
```diff
@@ -1,13 +1,187 @@
...
+  align-items: center;
+  flex-direction: column;
+  padding: 50px;
+  position: relative;
+}
+
+.form-question-3 {
+  max-width: 60vw;
+  display: flex;
+  flex-direction: column;
```
liaback-lowstett, line 44: Overall, super nice styling. Love the colors 🌈 . However, in question three it is a bit difficult to se how checkboxes and the text are grouped since the space between the text before and after the checkbox is the same.

liaback-lowstett (commented):
Awesome work! 🥇 I love the theme of the survey.

KarolCislowski (approved):
Great job!

You asked for a limitation of checked boxes. It is possible but there is no easy solution to it that I know. You need to store as a state the number of boxes that are checked and if it is equal or more than the max number you need to disable all unchecked ones.

PR #168, opened 2021-03-21
Project survey - Maria Pogosjan  

MariaPogosjan:
Link to my deployed project:
https://book-lover.netlify.app/

I think this was a fun project and I learned a lot about states. 

Vanilla javascript


ettem (approved):
Maria, first of all I think you have done an excellent job! I love your styling and also some features that increases the overall experience like the progress bar and that you only see one question at a time. Your code is clean and easy to read which is great. I wonder a bit about your naming of the components though; First/Second/ThirdQuestion is not so descriptive in comparison with Summary and NameQuestion. And also some comments in your code would be nice, for instance explaining a few of your functions. Other than that I have no comments. I love your app both subject wise and esthetic wise! Keep up the good work!

KarolCislowski (approved):
Great job!

PR #169, opened 2021-03-21
project-survey Elle Frankenberg

ElleFrankenberg:
Link to my deployed project:

https://epic-elion-e587ee.netlify.app

If I hade more time I would have done more with the styling 

it would have been good to have the structure explained from the beginning, for example where to put the useState to make the summery work.


code/src/components/Header.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const Header = () => {
+  return (
+  <>
+    <div className='logo-container'>
+      <img src="./images/logo.png" alt="logo"/>
+      <h1>Institute of Space Stuff (IoSS)</h1>
+      </div>
+      <div className='evacuation-plan-container'>
+      <h3>EVACUATION-PLAN-LEAVING-EARTH-123</h3>
```
IsabellaM5, line 11: I ran Lighthouse on your page and accessibility rating turned out quite high. One thing it showed though was regarding the use of h3 here on this line saying that heading elements are not in a sequentially descending order. Lighthouse also said that properly ordered headings that do not skip levels convey the semantic structure of the page, making it easier to navigate and understand when using assistive technologies. ⚙️

The way I understand Lighthouse’s message here is that you skipped an h2 heading and went straight to h3. One suggestion from my part would be to change the heading to an h2 here, that way you’re no longer skipping a heading. 🙂

It’s always good to run the page through different audit services such as Lighthouse and WAVE. Here’s also a checklist of things to test for in terms of accessibility: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility (see the bottom of the page).

code/src/components/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from'react'
+
+export const Footer = () => {
+  return (
+    <>
+      <h3>Approved for public release.</h3>
+      <h3>The electronic version is the official approved document.</h3>
+      <h3>(IoSS) does not take any responsebility in case of death.</h3>
```
IsabellaM5, line 8: This is outside of the requirements in my opinion, but one suggestion is to create some sort of container for these h3-elements to be in (a div for instance). Now they are pressed to the bottom of the page which isn’t that great in my opinion - the h3-elements needs some space to breathe as well 😅 

Also in the mobile view the h3-elements now hover over the text of the form since the footer is set to position fixed.. If you had a container for the footer this could be fixed with for example setting a background color 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
+import React, { useState } from 'react'
+
+import { UserName } from './Name'
+import { AgeRadioOption } from './Age'
```
IsabellaM5, line 4: I would consider the naming of the components here. For readability purposes, I would suggest you to name both the component file and the component the same name. So on line 3 for example, one suggestion is to name the component “Name” since that is the name of the js-file. 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+
+export const Form = () => {
+  const [ageGroupSpan, setAgeGroup] = useState('')
+  const [name, setName] = useState('')
+  const [item, setInputItem] = useState('')
+  const [planet, setPlanet] = useState('')
+  const [summary, setSummary] = useState(false)
```
IsabellaM5, line 17: One suggestion is to keep these useState in the order they are mounted in the code below. If you check out the states of them all in the React tools in the browser, they are in a different order from how they are displayed on the page. 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+import { AgeRadioOption } from './Age'
+import { SelectPlanet } from './SelectPlanet'
+import { BringItem } from './BringItem'
+import { Summary } from './Summery'
+
+
+
+
+export const Form = () => {
+  const [ageGroupSpan, setAgeGroup] = useState('')
```
IsabellaM5, line 13: I would consider the naming of the useState here. You set the state to “ageGroupSpan” but only named the set function to “setAgeGroup”. In the spirit of being consistent with naming states, a suggestion is to name the set function to “setAgeGroupSpan”. 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+  const [name, setName] = useState('')
+  const [item, setInputItem] = useState('')
```
IsabellaM5, line 15: Same comment really as in line 13. My suggestion is to either change name to “inputItem” or set function to “setItem”. 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+  const [planet, setPlanet] = useState('')
+  const [summary, setSummary] = useState(false)
+
+  const handelSubmit = event => {
+    event.preventDefault()
+    setSummary(true)
+  }
+
+ return (
```
IsabellaM5, line 24: As a general comment regarding the indentation of the code here in the return section, I would suggest you to use an auto formatter to get the correct indentation (2 spaces according to the guidelines). Good indentation is great for readability 🙂

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+  <>
+   {! summary ? (
+    <form className='form-container' onSubmit={handelSubmit}>
+      <div className='instruction-text-container'>
+        <p>To secure a spot in the evacuation progam.</p>
+        <p>Please fill in the form.</p>
+      </div>
+      <div>
```
IsabellaM5, line 32: Do you need to have a div around these mounted components in the form.js-file as well as in the respective component-files? 🧐 One suggestion is to remove the div’s from the form.js-file and only have them in the component-files.

code/src/components/Form.js
```diff
@@ -0,0 +1,71 @@
...
+        />
+      </div>
+      <div>
+        <BringItem
+        setInputItem={setInputItem}
+        item={item}
+         />
+      </div>
+      <div>
+        <button className='submit-button' type='submit'>Submit!</button>
```
IsabellaM5, line 57: I think it’s fine to have the submit button here since it’s not taking up that much space. Also you don’t have any other buttons on the page. In the spirit of reusability though, one suggestion is to put this in a component. 💪

code/src/components/Age.js
```diff
@@ -0,0 +1,33 @@
...
+
+const ageGroups = ['0-19','20-29','30+'] 
+
+export const AgeRadioOption = ({ageGroupSpan, setAgeGroup}) => {
+  
+  return (
+  <div className='age-input'>
+    <p>Age</p>
+      {ageGroups.map((age) => (
+        <label htmlFor='age' key={age}>
```
IsabellaM5, line 11: Since you’re wrapping the input tag in a label tag, you don’t need a htmlFor in the label tag. I suggest you remove this from the code since it’s not necessary 🙂

code/src/components/Age.js
```diff
@@ -0,0 +1,33 @@
...
+        <input
+        name='age'
+        type="radio"
+        value={age}
+        onChange={(event) => setAgeGroup(event.target.value)}
```
IsabellaM5, line 16: I would consider this function that you’ve put here directly on the onChange attribute a bit. This code is working and this is one way to do this. But in the spirit of consistency I suggest that you stick with one method on how to handle the event when the user is putting in a value here (by choosing a radio button) which you wrote differently in the other components (you made a separate function). 🧐

IsabellaM5 (approved):
You did a great job with this project! 🎉 🐬

You got the basic functionality in place in accordance with the blue level requirements. You also implemented some validation to the form, which is a red level requirement! Loved the idea of the form, haha! I’m all set then in case we’d need to evacuate earth 😅 

Your project is hereby approved! 👍 For some specific comments/suggestions, see the comments made in the code files.

/Isabella

KarolCislowski (approved):
Great Project! Remember to always create nice documentation in the readme file. 

PR #170, opened 2021-03-21
Project 10 - Darcy

darcywinterhill:
https://job-connect-form.netlify.app/

I had already managed to build a some what working project when the .map() was covered on the code session.
I got the mapping to work but it rendered one section per option.
So instead of having one section with the question 'In which one of our offices would you like to work?' and all the options underneath, I got 4 sections with height: 100vh under eachother with the question and one option. Didnt' have time to restructure.

I also tried to create a component with the OK button after I was done, which worked, and when I have more time I will restructure the site to have it look good, not only working.

I think the project's OK, but I had to redo some things when I deployed it and tested on my iPhone where a lot of stuff didn't quite work as they did in my desktop browser. After this the layout is not exactly how I wanted it to be.

I have read about the accessibility again but I'm still not sure what is best practice. If I don't add tabIndex='0' to my h:es and p:s I can't get the screen reader to read them, but I don't find any info about it being necessary to do tabIndex to elements like this.

I can't get the flexbox in the summary section to work when deployed and viewed on iPhone.


code/src/components/Welcome.js
```diff
@@ -0,0 +1,28 @@
...
+                <h1 className='welcomeText' tabIndex='0'>Let's connect!</h1>
+
+                <div className='welcomeContent'>
+                    <p tabIndex='0'>We would love to get to know you better.</p>
+                    <p tabIndex='0'>Tell us a bit more about what kind of job you're interested in and we'll get in touch!</p>
+                </div>
+            </div>
+
+            <button
+            className='startAndBackButton'
```
julialeonieros, line 16: Good naming, it's easy to understand what it does!

julialeonieros (approved):
Very nice looking site! I like the colors and overall design, and it works well 🎯  Nice with validation, that it isn’t possible to submit if all questions aren’t answered. A nice extra feature could be that it’s not possible to continue/click OK it the field for e-mail is empty. The code is clean and easy to follow Great job! 👏

PR #171, opened 2021-03-21
Week 10 project - Purchase order form

FD000:
Link to my deployed project:
https://kind-lamport-4a7b21.netlify.app/ 

This weeks project was fun and I sharpened my skills in React. But I face troubles when creating a logic to ask one question at each time and were the question disappears when answered and the next question pops up when clicking to the next button. 


code/src/components/Form.js
```diff
@@ -0,0 +1,115 @@
...
+    const [sumHidden, setsumHidden] = useState(true);
+    const [Size, setSize] = useState("");
+    const [myamount, setmyamount] = useState("");
+    const [Color, setColor] = useState();
+    const [myname, setmyname] = useState("");
+    const [myemail, setmyemail] = useState("");
+    const [wantsNewsletter, setwantsNewsletter] = useState("");
+    const [myphone, setmyphone] = useState("");
+    const [Rng, setRng] = useState("No Opinion");
+
```
MariaPogosjan, line 24: I see that you named some of the states with a capital letter, like Size and Color. Maybe it is better to be consistent and use lowercase.

code/src/components/Form.js
```diff
@@ -0,0 +1,115 @@
...
+        setwantsNewsletter(e.target.value)
+    }
+
+    const SelectEmail = (e) => {
+        console.log(e.target.value)
+        setmyemail(e.target.value)
+    } 
+
+    const SelectName = (e) => {
+        console.log(e.target.value)
```
MariaPogosjan, line 46: When you are done with the project it is better to delete the console.log :)

code/src/components/Form.js
```diff
@@ -0,0 +1,115 @@
...
+        setColor(e.target.value)
+    }
+
+    const OnSumChange = (event) => {
+        event.preventDefault()
+        setsumHidden (false)
+    }
+
+    if (sumHidden === true) {
```
MariaPogosjan, line 70: You have in the state declared sumHidden to "true". Therefore you could ad if (sumHidden) instead of if (sumHidden === true). Just a thought 👍

code/src/components/Wrapper.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+import Form from './Form'
+
+const Wrapper = () => {
```
MariaPogosjan, line 5: What is the difference between Form.js and Wrapper.js? I was thinking that it might be a good idea to add the Form.js directly into App.js and skip the wrapper component. But maybe there is a reason for it and that I don't have the knowledge to see it :)

code/src/components/Amount.js
```diff
@@ -0,0 +1,20 @@
+import React, {useState} from 'react'
```
MariaPogosjan, line 1: Here you add the {useState} but you are not adding any states here, only sending them through props from the Form.js. Therefore I would recommend not adding it in the import. I get a warning message in my console if I am not using something :)

code/src/components/Radio.js
```diff
@@ -0,0 +1,45 @@
...
+        {suitColors.map((group, imageIndex) => (
+          <label key={group} >
+            <div className="box-header">
+              <div className="box">
+                <div><Image image={images[imageIndex]} /></div>
+                <div className="container">
+                  <input
+                  type="radio"
+                  value={group}
+                  onChange={e => IselectColor(e)} // Du kan gära onChange ={SetColor}
```
MariaPogosjan, line 23: Don't forget to delete comments when you're done with your project ;)

code/src/components/RangeSlider.js
```diff
@@ -0,0 +1,35 @@
...
+            if (decideRange <= 25) {
+                return ( <p className="satisfaction">Bad: {decideRange}</p>)
+            } else if (decideRange > 25 && decideRange <50) {
+                return ( <p className="satisfaction">Not so good: {decideRange}</p>)
+            } else if (decideRange > 50 && decideRange <75) {
+                return (<p className="satisfaction">Good: {decideRange}</p>)
+            } else if (decideRange > 75 ) 
+            { return (<p className="satisfaction">Very good: {decideRange}</p>)
+            }
+    }
```
MariaPogosjan, line 14: Love your function here! I tried to do the same but didn't manage to do it correctly. Great job! 💯

code/src/components/Amount.js
```diff
@@ -0,0 +1,20 @@
...
+
+const Amount = ({choosedAmount, IselectAmount}) => {
```
MariaPogosjan, line 3: I love that you use the name of the value your sending instead of props. I am going to do the same for this week's project, super smart! 🥇

MariaPogosjan (approved):
Hi! 
I think your idea is so smart to do the survey as an online shop. Great thinking! Everything works as it should and you have done an amazing job with your React project. 

Your code is easy to read and follow. I would recommend you take one more look at your code before sending it so you don't miss deleting console.log and comments that don't give any specific information absout what the code is doing. 

Great job! 💯 🥇

PR #174, opened 2021-03-21
Project Survey - Mette

ettem:
Link to my deployed project:
https://gracious-hypatia-2912b1.netlify.app

Blue level is hard enough for me since javascript and react is still a struggle. So with that said, I guess it is ok but my projects will never be eyecandy.

I wonder why everything looks like sh*t in Safari for me always. Is there some secret mystery how to write code that complies with Safari, how do I make things look nice there as well?


code/src/Components/Summary.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+function refreshPage() {
+    window.location.reload(false);
+}
+
+const Summary = (props) => {
+    const { name } = props;
+    const { type } = props;
+    const { buildstyle } = props;
```
tuttibalutti, line 10: This way of using state and props is new to me. Looks neat😺

code/src/Components/SurveyRadio.js
```diff
@@ -0,0 +1,56 @@
...
+
+const SurveyRadio = (props) => {
+    const { type, setType } = props;
+
+    return (
+        <div className="radio-form" >
+            <p className="radio-form-headline">How would you prefer to live?</p>
+            <label>
+                <input className="radio-input"
+                id="radio"
```
tuttibalutti, line 11: If you want to associate the `label` element with the input element you should set a `for` attribute to the `label ` that has the same value as the `input id`🙋‍♀️  This link helped me https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

tuttibalutti (approved):
Mette🤩  Great work with this week's project! I learned a lot by looking at your code. I think it looks really neat and that it's easy to follow. I really liked the look of your survey, you have a good eye for design🧐  Keep up the good work 🥳  🙌

KarolCislowski (approved):
Great job!
Remember to inform your user if the form field is required 🤓 

PR #175, opened 2021-03-21
Week 10, Project survey - Isabella Markskog

IsabellaM5:
**Link to my deployed project**
https://quizzical-nightingale-6208b3.netlify.app/

**My reflections on how this week's project turned out**
I think this week's project turned out great! I did the blue level requirements and after that I proceeded to do the black level requirement of doing a multi step form and the red level requirement of adding validation to the form. 

At the start of the week I had a bit of code in place. Then I proceeded to read a lot of resources etc. instead of actually doing any coding to speak of for the next 2,5 days since I couldn't really decide what to do with my project. The main topics I was struggling to decide on was how to structure the flow of the components and how to implement the conditional rendering and functions to proceed onwards in the form. In the end I managed to decide what to do, and also added a bit of other functionalities, so I proceeded to then code away!

**Things I'd like to have clarified or explained in more detail**
Nothing specific at the moment. I want to continue to learn more about React and the different hooks you can use 😄


TobiasHagberg:
Hi Isabella!

It shows you had fun and enjoyed yourself this week. React is really powerful and you did a impressive job maneuvering the possibilities with the tools we have available so far. Great job! :)

TobiasHagberg (approved):
All and all this project really impressed me and I'm especially curious about how you went about mapping and planning it all out!? There are nesting of buttons that goes through all layers of the app and there some serious logic and reasoning going on and me personally would have needed a drawing/sketch :)

I would also like to hear more about creating the separate stages of the survey. The code is very well structured, the naming is good which makes it easy to follow and understand. If I where to be a team member of this project I would have applauded your code and wouldn't have any problems with continuing the work! Good job Isabella, looking forward to seeing what you create next :)

code/src/components/FirstPage.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import FormButtons from './FormButtons';
+
+const FirstPage = ({ step, setStep }) => {
+  return (
+    <>
```
KarolCislowski, line 7: Why did you used <></> if you have only one div inside?

code/src/components/Form.js
```diff
@@ -0,0 +1,82 @@
...
+              <SushiQuestion
+                sushi={formData.sushi}
+                callbackOnChange={handleInputChange}
+                step={step}
+                setStep={setStep}
+              />
+            )}
+
+            {step === 3 && (
+              <>
```
KarolCislowski, line 64: again react.fragment when you have only one element inside.

KarolCislowski (approved):
Great job!
Remember to use react fragments only when needed 🤓 

PR #178, opened 2021-03-21
Project 10 Survey

AsaHildebrand:
Link to my deployed project:
https://cat-survey.netlify.ap

I think this project turned out quite well, I have started to understand react and how it is working but i have been struggle a lot to get everything to work and I have spent many hours of the basic in React.


code/src/ColorQuestion.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react' 
+
+export const ColorQuestion = ({ color, onColorChange}) => {
```
TheErry, line 3: Another way to export (that you probably already know about) when you only have one function in the code to export is to use the un-named export `export default ColorQuestion` which could be placed at the end of the file instead. But this works too :)

code/src/index.css
```diff
@@ -1,13 +1,122 @@
...
+.summary {
+  display: flex;
+  flex-direction: column;
+  font-style: oblique;
+}
+
+.ending {
+  font-weight: 700;
+  height: 200px;
+}
```
TheErry, line 122: Good work with the classNames! It's easy to understand and where they take place. I have no other comments on the CSS.

TheErry (approved):
Hi Åsa! I really like your idea for this project! It's cute and funny, and well made :) The code runs well and I couldn't find any bugs.
I don't have a lot to comment in the code. I really like the naming of classes and functions and the code is very easy to read There are however some structure of the props that isn't consistent that you might want to take a look at. Otherwise it seems very clear to me, and I like how you divided the components. The responsiveness seems fine and it is also good on an accessibility point. Perhaps the grey paw print could be slightly lighter, because the smaller black text disappears a little bit in it in the summary. But there's no major thing. I'm gonna put this on approve and hope the teachers agree with me ;) (I am however running a fever again today so I'm a bit dizzy and might have overlooked something...) Great job on the project!

KarolCislowski (approved):
Great job!

PR #179, opened 2021-03-21
Anna Lenksjö Project Survey

annalenksjo:
Link to my deployed project: https://condescending-hawking-aea228.netlify.app 

My reflections on how this week's project turned out: I got stuck a long time on how to construct the summary part. This prevented me from doing more styling and a more advanced survey which would have been fun. But I guess I just have to go back to the project in that case. 

We talked a bit in our group about that we would like to have concepts more explained in like powerpoint more visual ways before looking at the code. It would make it easier to understand the big picture.


code/src/App.js
```diff
@@ -1,9 +1,58 @@
...
+  const [temperature, setTemperature] = useState()
+  const [summary, setSummary] = useState(false)
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
+     <>
+     {!summary ? (
```
sirimatsdotter, line 20: Nice with a ternary operator! :)

code/src/components/Input1.js
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <div>
+      <h3>
+        Favourite vacation activity 🏄
+      </h3>
+      <label>
+        Type:
+        <input 
+          type="text"
+          onChange={(event) => {onActivityChange(event)}}
```
sirimatsdotter, line 18: To simplify the code you should be able to only write onChange={onActivityChange} - since event is the default parameter so you shouldn't need to specify this.

code/src/components/Input3.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+
+const optionsQuestionThree = ['< 20', '20-25', '25 <']
```
sirimatsdotter, line 3: I think it could have been nice to write 25 + instead of 25 < since this is not something most non-coders might use in everyday life when talking about temperatures.

code/src/components/Submit.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+export const Submit = ({setSummary}) => {
+
+  const formSubmit = event => {
+    event.preventDefault()
+    setSummary(true)
```
sirimatsdotter, line 7: Currently, one can still press submit even if not filled in a single input, perhaps you should implement a way to not allow the button the send to the summary until all inputs are filled in

sirimatsdotter (approved):
Great work Anna! I think the survey is really nice and you nailed all of the blue requirements!

If you want to improve the project I've added some comments related to user friendliness and validation!

KarolCislowski (approved):
Great job! 

PR #187, opened 2021-03-21
Project 10 Maidelin Rubio

rubmaigut:
Link to my deployed project:


FD000 (commented):
Fantastic work, I was very impressed by your code. One of the most beautiful project I have ever seen. 

I would like you to review 1 thing in your project:

1) When you are answering questions, there is a counter in the bottom that informs you what question you are at, that one counts wrong. For instance when you have answered first question, it says "3 answered of 5".

rubmaigut:
> Fantastic work, I was very impressed by your code. One of the most beautiful project I have ever seen.
> 
> I would like you to review 1 thing in your project:
> 
> 1. When you are answering questions, there is a counter in the bottom that informs you what question you are at, that one counts wrong. For instance when you have answered the first question, it says "3 answered of 5".

Hi @ FD000, when you answered the first question as "yes", the code avoids the protein question because all the answers are "non-vegan" options. So if your answer is "yes", the remaining questions are 3 and you just choose 3 more options to see the results. when your answer is "No", you need to choose 4 options more but always the length of the array is 5 questions.

code/src/Data.json
```diff
@@ -0,0 +1,69 @@
...
+        "lamb",
+        "meat",
+        "Pork"
+      ]
+    },
+    {
+      "id": 2,
+      "question": "Exclude one of the following... ",
+      "filter": "health",
+      "optionType": "radio",
```
KarolCislowski, line 31: I think the checkbox would be a better option here. 🤓

KarolCislowski (approved):
It is an awesome project! 
And now I need to go to eat something because it made me hungry 🥘

PR #188, opened 2021-03-21
Survey - Astrid Persson

Aepera:
Link to my deployed project:


darcywinterhill (approved):
Great job with your survey!✨
It's easy to follow and the functions you have are working as they should.
If you'll ever return to the project, a functioning validation is closer than you think 😉

KarolCislowski (approved):
Great job! 
Extra points for 🥥 

PR #189, opened 2021-03-21
[Not yet finished] survey project, or in my case a Dungeon World playbook recommendation quiz

Martno96:
Link to my deployed project:
heuristic-ramanujan-44af52.netlify.app

I made the mistake of going crazy with the scope and vastly underestimating how much more time it takes to write something slightly more complicated like my playbook filtering idea. Next week I will definitively start by fulfilling the base requirements, then make something more fun out of it from there.

This project is very far from finished and I haven't had time to tidy up my code :( sorry about that. I hope the code is still interesting to review and that I can get some feedback of the (albeit overambitious) functions that are in place :)


code/src/Components/RadioButton.js
```diff
@@ -0,0 +1,20 @@
...
+            <input
+                name="radio" 
+                id="radio-button" 
+                type="radio"
+                checked={props.source === buttonValue}
+                onChange={() => {props.toChange(buttonValue)}}                       
+            ></input>
+            <label htmlFor="radio-button">
+                {props.question.options[props.optionNumber].label}
+            </label>
```
klimenko3007, line 15: From what I can see, all of your radio-buttons have the same id and label. So, when you click on any label, only first button is checked. Radio buttons should have unique ids and labels should correspond to the ids.

code/src/Components/RadioButtonsWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+                <RadioButton 
+                    question={props.question}
+                    toChange={props.toChange}
+                    optionNumber={3}
+                    source={props.source}
+                />
+            </div>
+        </form>
+    )
+}
```
klimenko3007, line 41: I was wondering, whether you could think of a way of creating buttons with a map method rather than creating the same component over and over again. in you json file you have object called question that has an array of options. You can loop through that array and create radio button for each options.label. And of course each button should receive the same name but different ids.

code/src/Components/NavButton.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+const NavButton = (props) => {
+    
+    return (
+        <button 
+            className="nav-button"
+            onClick={() => props.navigate(props.direction)}
+        >{props.direction === -1 ? "<" : ">"}</button>
+    )                           //this might fail cuz strings??
```
klimenko3007, line 10: Love the ternary operator!!! I could not come up with one for the button's direction. Mine only go one way

code/src/Components/NavWrapper.js
```diff
@@ -0,0 +1,26 @@
...
+            <PageCounter 
+                currentPage={props.currentPage}
+                lastPage={props.lastPage}
+            />
+            <NavButton 
+                direction={1}
+                navigate={props.navigate}
+            />
+        </div>
+    )
```
klimenko3007, line 22: This is very clean and structured code!

code/src/Components/OptionsWrapper.js
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+        return (
+            <div className="question">
+                What kinda question is this?
+            </div>
+        )
+    }
+}
+
+export default OptionsWrapper;
```
klimenko3007, line 53: Do I understand it correctly, that this is the only conditional rendering that is actually used at the moment. In App.js you have several but I do not see that they are actually being used.. Perhaps, I am just confused there.

klimenko3007 (changes requested):
Hi, Martin! I think you have a really awesome idea of the playbook and I remember explaining it during the demo. So if you implement it, it will be fantastic! 
But as you know yourself, the project is not quite finished. I suggest, perhaps, you first complete the basics (as you already noted yourself) just make your questions of the survey work smoothly and then add you cool extra features one at the time. 
I can see that the states you are using are changing depending on the choice, but still would be great to see the actual summary of answers.

Guiding your user through the process would be useful too. I am sure you have already thought about adding text but extra instructions would be good. Like in the selector, it is not right away clear why text flashes in red color. I have figured it out eventually, but not at first glance. And the squares that turn blue look a bit like buttons and I was really tempted to click on them (hehe). Perhaps they will be buttons at the end ;) 
So, to sum up, I request changes to give you chance to finish your really cool project idea :D

KarolCislowski (changes requested):
It looks great so far!

KarolCislowski (approved):
🥇

code/src/Components/RadioButtonsWrapper.js
Martno96: That is a great idea! Thank you ^^ will fix in a future commit

PR #193, opened 2021-03-21
React-Survey-SandraWallén

sandra-wallen:
https://practical-williams-b581cf.netlify.app/

I had a really great time with this weeks project. I'm really pleased with my code and I feel that I'm starting to understand React and the perks of it.

I have a component called Progressbar that is not mounted anywhere, it is still work in progress.

I'm aware that we should style for mobile and then add media queries for tablet and desktop, unfortunately I did all the styling for desktop and had to do the media queries for mobile instead.


Yennie-Leek (changes requested):
Hey Sandra! Its a really good and fun project you've done💫 
Your code is easy to read and clean. Good job on that! 
And also Im impress with how you think one step ahead and create less code but with the same or more functionalities. 

The only feedback I have is to look over the site and make it a bit more accessible as it was one of the requirements. I tried to tab my way through and I couldn't enter the start button or move forward after that. 

With that said, you are really getting this and I take my hat off! 🎩 
Great job!🥳

Yennie

KarolCislowski (changes requested):
Awesome project! 
But panda Karol that loves wearing hats and prefers to play video games rather than eating apples will love to use a tab to navigate in your form :D 
Make 🐼 happier, please 😸

KarolCislowski (approved):
😻

PR #197, opened 2021-03-21
Project Survey - Erika A Porath

TheErry:
Link to my deployed project: https://swap-card.netlify.app/

I had a bit chaos this week because first I was too sick to manage to do anything. And when I felt better, the only computer I have that can handle node.js crashed... So I got a new computer on Friday and had that one set up then. Then I made this project during the weekend. Depending on that I had very short time I think this project went very well. There are things I would like to have add and changed if I had more time. I cheated a bit with the responsiveness, because I made a card that looks the same in all screens. It looks nice and kinda as I had in mind though. 

I hade the advantage that a lot of people had finished this project before me, so I could look at their code to get some tips. I used the same approach as one of my team-mates because I thought it was very clean and logic. But now it feels like I've copied a lot... I understand every piece of code I put in though, but I'm not sure if I would be able to replicate it if I had to though. And I don't know having all data in a json-file instead of the components count as it being reusable either. I liked that approach.


code/src/components/SubmitButton.js
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+
+const SubmitButton = (props) => {
+  const { setSummary, isRequired } = props
+  const handleSubmit = () => {
+    setSummary(isRequired())
+  }
+
+  if (isRequired()) {
```
Polkapetunian, line 9: Good job of  enabling and disabling submit button!

code/src/components/SwapCard.js
```diff
@@ -0,0 +1,33 @@
...
+        <h1 className="swap-card-title">
+          I'm looking for a swap-buddy!
+        </h1>
+      </div>
+      <div className="information">
+        <p>My name is: <span>{props.name}</span></p>
+        <p>I'm coming from: <span>{props.country}</span></p>
+        <p>I'm looking for a buddy in: <span>{props.region}</span></p>
+        <p>My swap criteria is: <span>{props.amount}{props.unit}</span></p>
+        <p>I have these allergies/preferences: <span>{props.allergies}</span></p>
```
Polkapetunian, line 23: It's a bit confusing mixing allergies and preferences. I am allergic to Nuts or do I prefer Nuts? Dividing this in Two different input is a good idea.

code/src/components/SwapCard.js
```diff
@@ -0,0 +1,33 @@
...
+        <p>Other information: <span>{props.other}</span></p>
+      </div>
+      {/* <div className="image-container">
```
Polkapetunian, line 26: Either erase this that you've commented out or try again to make it work.

code/src/components/CheckboxInput.js
```diff
@@ -0,0 +1,34 @@
...
+const CheckboxInput = ({ choice, setChoice, data }) => {
+  const handleChange = (event) => {
+    choice.includes(event.target.value)
+      ? setChoice(
+        choice.filter((item) => item !== event.target.value)
+      )
+    : setChoice([event.target.value, choice])
+  }
+  return (
+    <div className="check-input">
```
Polkapetunian, line 12: I know checkboxes are hard to make so good that you made them. There is however an error: if you check something and get the result and then you press Again?-button to make another one, the checkboxes do not look checked but the previously selected value sticks and the new values you check are just added on to that.

code/src/App.js
```diff
@@ -1,9 +1,113 @@
...
+          setSummary={setSummary}
+        />
+      </div>
+    )
+  } else {
+    return (
+      <div className="form-container">
+        <form>
+          <Heading/>
+          <TextInput
```
Polkapetunian, line 68: Using a json-file is probably efficient if you want to make changes but it's a bit harder for me to wrap my head around. Could there be an issue with this method that creates double labels fort the inputs? Beacuse in WAVE I see errors that there are two labels associated with the same input.

Polkapetunian (changes requested):
I think you did a good job in the short amount of time you had due to computer failures. However you need to address the accessability to meet the blue goals. Look at the site in WAVE https://wave.webaim.org/ and you will see some errors that need fixing. One is double labels and one is bad contrast text (the text that says "not required"). You can test contrast in this tool https://webaim.org/resources/contrastchecker/  I think the font size should also be increased, the smallest characters are a bit two small using this font.

code/src/components/ReturnButton.js
```diff
@@ -0,0 +1,18 @@
...
+    setSummary(false)
+  }
+  return (
+    <button 
+    className="return-button"
+    onClick={handleReturn}
+    >
+      Again?
+    </button>
+  )
```
KarolCislowski, line 15: it would be nice if your inputs would be reset after clicking it.

KarolCislowski (approved):
Great job!
For the next step, it would be nice to try separate questions to multiple pages. 📖

PR #228, opened 2021-10-22
Health and Wellness Assessment App by Lousanne

loulunds:
[Health and Wellness Assessment App](https://health-and-wellness-assessment.netlify.app/)

This week was so great. I learned a lot from this project. I realized that you need to sometimes check how to do things in React  but anyhow things just worked! 

My validation only does not allow the user to proceed if the input field is empty and if the user types a wrong input type ex. letter instead of numbers in measurements.


Asivol93 (approved):
Lousanne! Your survey is amazing and your code is complex yet easy to follow, you should be super proud :) And it passes in an accessability tool!

## **🔵  Blue Level (Minimum Requirements)**

- Your survey consist of at least 3 questions.
- At least one question  is with radio buttons.
- At least one question should be with a select dropdown. (This is missing but I don't know how important it is!)
- There is a submit button. When pressed your app hides the input form and shows a summary of the user's submissions.
- Your site does follow accessibility guidelines

PR #234, opened 2021-10-22
Elin Diczfalusy week 10 Project survey "vabb-life"

Ajliin:
Link to my deployed project:
https://project-survey-ajliin.netlify.app/#firstPage

Im happy with the project, I like the styling and that I had time to try out new type of inputs like slide range and that I solved a progressbar


code/src/index.css
```diff
@@ -1,13 +1,168 @@
...
+
+* {
+  box-sizing: border-box;
+}
 body {
   margin: 0;
-  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen",
-    "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue",
-    sans-serif;
+  font-family: raleway;
```
hemmahosjessi, line 15: Maybe add this one on the other font-family-row? Not sure if we should delete the others or keep them. I think it is good to have a fallback font. But not sure how that works. Do you know?

hemmahosjessi (commented):
Elin, so nice work! Love that you made your own design, and it looks really good! ⭐️
The separate components, with their own css file is so inspiring to look at. The code is tidy and easy to follow.  Like that you write comments in the css.  
Seems like you really understand what you are doing, not only looking at this, but when hearing you speak about the code.  See you! /Jessi 🦊

PR #247, opened 2021-10-23
Project-Survey by Madelene Trang

MT-dotse:
# Survey form with React

The aim of project was to learn how implement "useState"-hooks and use form fields in React, as well as how to connect HTML form input to state. I made a survey imitating a cinema check out experience, asking the user to pick movie, seats, dates and snacks. 

I enjoyed working and learning about "useState"-hooks, and would like to challenge myself to use more conditional renderings in future projects. 

## View it live
https://mt-dotse-scarymovie-survey.netlify.app/


Alisebrink (commented):
Hi Madelene!
I have taken a look at your code and it seems to be working fine and it is understandable and easy to follow along. One thing I have to comment on is that it would be easier to understand and follow along if you make some more comments in it. That would make it easier for the one making the review and for yourself later on to understand it. 

Also I think that your images are to big to use on a website. Every time I go onto a question it takes up to 2 seconds to load the image. I checked the sizes of your images and they were up to 3.32mb which is way too big for a webpage. If a person with a slow internet connection were to use your website it would take a long time for them to load the images. I have a fast connection and it takes me 2 seconds to load them. For your next project I would recommend you to use images that are not bigger than 500kb. To make this happen you can either crop them using [https://photopea.com](https://photopea.com) or [https://tinypng.com](https://tinypng.com)

code/src/components/IntroPage.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const IntroPage = ({ onStepChange }) => {
+	return (
+		<div className="welcome-message-container" tabIndex="0" id="first-question">
+			<h1 className="welcome-message">
+				Want to watch a horror movie?....................{' '}
+			</h1>
+
+			<button className="button" type="button" onClick={onStepChange}>
```
Alisebrink, line 10: As for one who doesn't like scary movies I would love to have the option to say no :D then maybe you could use and if/else and show not so scary movies for those who wants them.

code/src/components/Date.js
```diff
@@ -0,0 +1,38 @@
...
+const Date = ({ date, onDateInputChange, onStepChange, onMinusStepChange }) => {
+	// const { nameInput, onNameInputChange } = props;
+
+	return (
+		<form className="form-container5" tabIndex="4">
+			<label htmlFor="dateInput">
+				<h1>Pick a Date</h1>
+			</label>
+
+			<input
```
Alisebrink, line 12: I really like the fact that you used a date picker as an input! This would also be a great place to have an if/else statement. You could after picking a date show different seats depending on which date they picked or show different movies depending on which date they pick.

PR #285, opened 2022-03-16
Workplace Flexibility Survey - Savannah Hayes

savannah-hayes:
Link to my deployed project:

Workplace Flexibility Survey By Savannah Hayes
https://clever-cray-b0187c.netlify.app/

I am happy with the way this project turned out because I learned a lot about building custom form elements
and adding form validations this week. Working with React Hooks is something I really enjoy because I really
enjoy working with React!

I would like to have component explained in more details. More specifically when you should create a new 
component and how to use them in the right way to keep your code dry.


bostrombundock:
Savannah, your code is interesting to read and easy follow and the moving stars are great! This review is maybe not taking you forward, but I really think your css.js is easy to follow and it makes sense even if I just read your code once. Inspiring!

PR #289, opened 2022-03-18
Customized invitation card - Emma Högberg

emmahogberg88:
https://2kidsparty.netlify.app/


qwer-ctrl (commented):
I love your idea, I would definitely use this kind of app for kids invitation cards. As a user I like most that form doesn't delete all information if I start over so I can test different styles without much effort. Also it is nice that current time is selected in dropdown menu.

Code was very clear with good structure like always so it was easy to follow. Some things was little advance for me so it takes time to understand it. I think styling for progress line with if else statement was something to learn from you :)

PR #290, opened 2022-03-18
project-survey Ida Halling 

id4h4lling:
Link to my deployed project:
https://lifeonplanetearth-survey.netlify.app/

It went pretty good at the end I feel that I get React more and more, if I hade more time I would have put more effort in to validation. Also on the styling of radio-buttons. 

The naming convention could be set to the question instead of the function, will change or not do that on future projects.


isabellwastfelt (commented):
Nice survey! 👍  
The code looks good and I really like the animation and how you written the questions.
I kinda miss a restart button in the end so I could to the survey again.
But overall good code that was easy to follow and understand 🌷

PR #291, opened 2022-03-18
Week 10 - Sofie Pellegrini 

SofiePellegrini:
Link to my deployed project:


code/src/index.css
```diff
@@ -1,13 +1,191 @@
...
+    width: 50vw;
+  }
+
+  .btn {
+    width: 10vw;
+    margin-top: 30px;
+  }
+
+  .btn:hover {
+    box-shadow: 0 0px 50px 0 rgb(255, 255, 255), 0 0px 50px 0 rgb(255, 255, 255);
```
EmmaaBerg, line 189: Nice detail of the buttons, matches the theme really well :D

code/src/App.js
```diff
@@ -1,9 +1,10 @@
...
+import Form from 'Components/Form'
 
 export const App = () => {
   return (
     <div>
-      Find me in src/app.js!
+     <Form />
     </div>
   )
 }
```
EmmaaBerg, line 10: You have a very clean entry point which is good!

code/src/Components/ThirdQuestion.js
```diff
@@ -0,0 +1,64 @@
...
+            <label className='dormitory'>
+              <input
+                type='radio'
+                name='dormitorys'
+                value='Slytherin'
+                onChange={onHomeInputChange}
+                defaultChecked={homeInput === 'Slytherin'}
+              />
+              Slytherin
+            </label>
```
EmmaaBerg, line 19: A suggestion for projects in the future is to create an array with the different options and then map through them, in that way you don't have to repeat yourself so much :D Something like
`const homeArray = ['Slytherin', 'Ravenclaw', 'Gryffindor', 'Hufflepuff']` and then use `{homeArray.map(() => (
))}` to go through the array.

EmmaaBerg (approved):
Really good work with this project. Your survey fulfills the requirements and I think it was easy to follow with the different components. The code follows the guidelines on how to write good code as well and is responsive as far as I can see. Left a comment on the radio buttons for future projects about DRY, but that's just a detail, and the more we learn about code the more natural it will be I guess :D

code/src/index.css
```diff
@@ -1,13 +1,191 @@
...
+.slider::-moz-range-thumb {
+  width: 25px;
+  height: 25px;
+  border-radius: 50%;
+  background: url(assets/harryslider.png);
+  cursor: pointer;
+}
+
+.input[type='range']::ms-fill-lower,
+.input[type='range']::ms-fill-upper {
```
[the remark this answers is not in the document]
SofiePellegrini: When I tested my project in different browsers I noticed that it looked really of in Explorer so these are for styling it in Explorer. Even though it seems impossible to get it completely right there.. :)

code/src/Components/ThirdQuestion.js
SofiePellegrini: Thanks for the tip! Didn't think about that :D

PR #293, opened 2022-03-18
project-survey Frida Axelsson

Nosslexa:
Link to my deployed project:
https://pensive-heyrovsky-44b016.netlify.app/


code/src/components/Summary.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import Sunflower from "icons/sunflower.png";
+
+const Summary = ({ email, name, ageGroup, location }) => {
```
EmmaSprings, line 4: I'm just got a tip in my code review,  that this way to destructing code can be used in the other components as well, instead of using the "const { propName, anotherPropName } = props", just to be consistent. So I thought I'd forward that too! :)

code/src/components/AgeRadioBtn.js
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div>
+      <form onSubmit={onStepChange}>
+      <h1>
+        {" "}
+        Choose your age group and get the opportunity to meet other gardening
+        enthusiasts in your age
+      </h1>
+        Answer:
+        {ageGroups.map((ageGroup) => (
```
EmmaSprings, line 16: I like this solution! To map an array.

code/src/components/IntroPage.js
```diff
@@ -0,0 +1,26 @@
...
+            We are so happy you want to join our gardening course this spring!
+          </h2>
+          <p>Enter your details in the following form to save your spot.</p>
+          <button className="btn" type="submit">
+            Sign up
+          </button>
+        </form>
+      </div>
+      <div className="icon-container">
+        <img src={Sunflower} className="icon" alt="Sunflower" />
```
EmmaSprings, line 21: Also like this solution to insert an image this way!

EmmaSprings (commented):
Well done with your survey Frida! Your site is responsive and meets the requirements. Nice job with conditional rendering too! I added a comment on your Summary component to forward a tip that I got myself, to use the same way to add props in all components, just for consistency. However your site works fine and is well functioning. Great work!

PR #294, opened 2022-03-18
Lisas Wine & Dine survey

lisabergstrom:
Link to my deployed project:

https://wineanddine.netlify.app/


code/src/Components/Buttons.js
```diff
@@ -0,0 +1,48 @@
...
+    e.preventDefault();
+    setSection(5);
+  };
+
+  const nextSection = (e) => {
+    e.preventDefault();
+    setSection(section + 1);
+  };
+
+  const previousSection = (e) => {
```
SofiePellegrini, line 14: Nice with the "back" button!

code/src/Components/Day.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+
+const Day = ({ setDay }) => {
+  const dayArray = ["Friday", "Saturday", "Sunday"];
+
+  return (
+    <div className="survey-item day">
+      <p className="section-heading">What day will you be joining us?</p>
+
+      {dayArray.map((item) => (
```
SofiePellegrini, line 10: Great with mapping the array instead of writing the code for each day.

code/src/Components/Options.js
```diff
@@ -0,0 +1,26 @@
...
+      </label>
+      <select
+        className="optionCity"
+        onChange={(e) => {
+          setOptions(e.target.value);
+        }}
+        id="option"
+        value={options}
+      >
+        <option disabled>Choose below</option>
```
SofiePellegrini, line 17: I noticed that even though you set `disabled` here you can still choose "choose below" as an option. I found this that maybe can work instead: `disabled={true}`.

SofiePellegrini (commented):
Great job Lisa! You fulfilled all the blue requirements. Also I love the theme :D

PR #300, opened 2022-03-19
Week 10 – Survey – Emma Lindell

emmajosefina:
Link to my deployed project: https://survey-project-emma.netlify.app/

My reflections on how this week's project turned out: So happy about how this week's project turned out. Also I didn't have any larger issues which was really nice.

Things I'd like to have clarified or explained in more detail: Nope, don't think so :)


JustineZwiazek (commented):
I enjoyed both your demo last Friday, and reading the code! Such a great job with this one, looks very professional as usual! Two things I wanted to highlight is the split of first and last name - this was thoughtful and added to the professional look, and the styling of the input boxes. I will take some inspiration from your code as I struggled styling select menu. For advice, the only thing I could think of is that you could use email validation, at the moment it is possible to add plain text as email. Super nice job and I can't wait to see more from you :)

PR #301, opened 2022-03-19
All update have done

eyahya-khan:
Link to my deployed project:


code/src/App.css
```diff
@@ -0,0 +1,225 @@
...
+    text-align: center;
+    cursor: pointer;
+    z-index: 99;
+  }
+  
+  h1,
+  h2,
+  h3 {
+    margin: 10px;
+  }
```
vanhaj, line 35: This is a great way of reusing the element for the same purpose!

code/src/Appsinglepage.js
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from 'react'
```
vanhaj, line 1: A suggestion for naming the components is using PascalCase, so for example instead of Appsinglepage it should be AppSinglePage. You can read more about it here: 
[https://www.notion.so/Guidelines-for-how-to-write-good-code-6a88a79b525f407f9bb039ea047bb7d2](url)

code/src/Appsinglepage.js
```diff
@@ -0,0 +1,80 @@
...
+import 'App.css'
+
+export const App = () => {
```
vanhaj, line 4: Another suggestion is also to name the component the same as the file, so that it doesn't get confusing! :)

code/src/Appsinglepage.js
```diff
@@ -0,0 +1,80 @@
...
+  const [name, setName] = useState('');
+  const [email, setEmail] = useState('')
+  const [number, setNumber] = useState('');
+  const [age, setAge] = useState();
+  const [showSummary, setShowSummary] = useState (false);
```
vanhaj, line 9: The naming here makes it really easy to follow along with the code further down! Good job!

code/src/components/Survey.js
```diff
@@ -0,0 +1,106 @@
...
+    const [personNumber,setPersonNumber] = useState('')
+    const [ageGroup,setAgeGroup]         = useState('')
+    const [email,setEmail]               = useState('')
+
+    const stepCounts = [
+      { id: 1},
+      { id: 2},
+      { id: 3},
+      { id: 4},
+   ]
```
vanhaj, line 26: This taught me about how to add the steps. I really appreciate this add on!

code/src/components/survey-components/Secondquestion.js
```diff
@@ -0,0 +1,66 @@
...
+        <label htmlFor="ageGroup" key={agegroups} className='agegroup-label' tabIndex={0}>
+        <input
+            className='agegroup-input' 
+            id="ageGroup"
+            type = 'radio' 
+            value={agegroups}
+            onChange={handleAgegroup} 
+            checked={props.ageGroup === agegroups}
+            tabIndex={0}
+        />
```
vanhaj, line 36: I could be wrong but if you move the htmlFor='ageGroup' to inside <input /> it will also be possible to click on the text next to the button and the button will be ticked in, instead of only the radio button. Let me know if I exaplained it wrong here, we could talk about it more :)

vanhaj (approved):
Really nicely done with this weeks project!
It's such a cool effect with the color change. Unfortunately I noticed one thing about that part, in mobile version it seems to not be working. Not sure if it was meant to be like that!

I also apperciate the fact that you added the required for the next button! A small detail but makes a big difference in my opionion, nice work!

Overall your project looks really good and it has been very educational for me to follow the code! Thank you :)

PR #303, opened 2022-03-19
Supper Society Survey - Arianna

ariallahyar:
Link to my deployed project: https://supper-society-survey.netlify.app/

My reflections on how this week's project turned out: I wanted to try to build a form that can be used for many contexts, and I managed to do that. This means that the theme can change, the number of questions can easily be increased or decreased, and the question types can be changed all in one place. When the data object is updated, the rest of the configuration is dynamic.

Things I'd like to have clarified or explained in more detail: I'm still struggling to get a grip on when to break code into a component, what to have at the entry component, and how to properly add global styles.


code/src/components/Button.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import "./Button.css";
```
rawisou, line 2: I like that you have a css file for each component and that you imported it in each component.
I have done the same, but I imported everything in the [index.js](https://github.com/rawisou/project-survey/blob/master/code/src/index.js) file! 

Looking at your code, this makes much more sense!

I asked Maksy just to be sure there're no big advantages or disadvantages between these two approaches (like with the compile and everything) and he said no! Both should work but your approach makes it clearer for others that there's a css file for the component. :) Great job!!

rawisou (commented):
Hi Arianna,

Your site works as it should and is responsive. It has fulfilled the requirements.
Your code is tidy and easy to follow. It appears that you have varied the techniques and approaches used in your code and I have learned a lot from this.

Great job!

PR #307, opened 2022-03-20
w10 - Jin Cho

jiiinCho:
Link to my deployed project: https://technigo-survey-2022.netlify.app

My reflections on how this week's project turned out:  The challenge was to structure each component to be re-usable, however, it is not achieved in this project because each question requires different input format. This can be improved in the future. Another challenge was including form validation. For example, question about author has alert state to avoid submitting in empty input value.


code/src/App.js
```diff
@@ -1,9 +1,23 @@
-import React from 'react'
+import React, { useState } from "react";
+import QuestionAboutName from "./components/QuestionAboutName";
+import QuestionAboutPainting from "./components/QuestionAboutPainting";
+import QuestionAboutAuthor from "./components/QuestionAboutAuthor";
+import Summary from "./components/Summary";
 
 export const App = () => {
+  const [answers, setAnswers] = useState([]);
+  const [count, setCount] = useState(0);
```
faypi, line 9: you could maybe use a more descriptive name like step or something like that :)

code/src/App.js
```diff
@@ -1,9 +1,23 @@
...
+  const onFormSubmit = (value) => {
+    answers.push(value);
```
faypi, line 11: I think the answers array is used to hold both answers and results of your validation which is a bit confusing to me, but it's also smart thinking! Good job :D

code/src/App.js
```diff
@@ -1,9 +1,23 @@
...
+  };
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <>
+      {count === 0 && <QuestionAboutName onFormSubmit={onFormSubmit} />}
```
faypi, line 17: _**onFormSubmit**_ I think is not the best name for this function because it sounds like a form event listener but it's a method that is used after the actual form has been submitted.

code/src/components/QuestionAboutAuthor.jsx
```diff
@@ -0,0 +1,70 @@
...
+    "Julie",
+    "Joanne",
+  ];
+
+  const onAlert = () => {
+    setAlert(true);
+    setTimeout(() => {
+      setAlert(false);
+    }, 2000);
+  };
```
faypi, line 20: How is this alert being used? I tried to trigger it on my browser but couldn't get it to work

code/src/components/QuestionAboutAuthor.jsx
```diff
@@ -0,0 +1,70 @@
...
+      </p>
+      <img
+        className="question-img"
+        alt="JK Rowling portrait"
+        src="https://www.barnensbibliotek.se/portals/0/image/RowlingJK_Foto_Debra%20Hurford%20Brown_460q.jpg?ver=IB4sUphdL4RjieEQ5OyONg%3D%3D"
+      />
+      <select
+        className="question-author"
+        onChange={(e) => setSelect(e.target.value)}
+        required
```
faypi, line 53: Nice - I forgot to add HTML validation on mine heheh

code/src/components/Summary.jsx
```diff
@@ -0,0 +1,20 @@
...
+    <p className="summary-text">
+      Your first answer about <strong>The persistence of memory</strong> is
+      <span className="summary-answer">
+        <strong> {result[1]}</strong>
+      </span>
+    </p>
+    <p className="summary-text">
+      Your first answer about <strong>JW Rowling</strong> is
+      <span className="summary-answer">
+        <strong> {result[2]}</strong>
```
faypi, line 15: Maybe you would like to add the actual answer along with the wrong/correct result :)

code/src/index.css
```diff
@@ -5,9 +5,116 @@ body {
...
+  }
+  .question-img {
+    max-width: 360px;
+    height: auto;
+  }
+
+  .submit-button {
+    max-width: 360px;
+  }
+}
```
faypi, line 120: Nice and very clean CSS rules! Good job!

faypi (commented):
Good job! Very inspiring approach to the project 👍

PR #308, opened 2022-03-20
Project survey

rawisou:
Link to my deployed project: https://libertasbookclub.netlify.app/


JaEngd (approved):
Hi Rawi!

Good job making this survey, it looks and works out great! 

I really like the styling with the choice of colors, composition, text and forms. The functionalities works well and and I like that you've added a function that automatically redirects you to the next page. I also ran your page trough an accessibility tool and it has no errors.

You've written a clean and readable code and maintained consistency on all files and components. For example that you've added one css file per component and that the elements in the input tags are on different rows.

I'll dig deeper to see if I can provide any useful feedback. 

Again, great work Rawi! :)

All the best,
Jacob

PR #321, opened 2022-03-20
Week 10 project

JensBergqvist:
Link to my deployed project:
https://elegant-blackwell-b5f79c.netlify.app/

### DIFF
--- a/code/src/components/Name.js
+++ b/code/src/components/Name.js
@@ -9,7 +9,7 @@ import React from 'react'
     return (
         <section className="container">
           
-        <h2>Alright let's go!</h2>
+        <h2>Thank you for participating!</h2>
         <label htmlFor="name">Please type your name</label>
         
         <input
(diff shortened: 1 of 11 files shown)


code/src/components/Name.js
```diff
@@ -0,0 +1,29 @@
...
+          
+        <h2>Thank you for participating!</h2>
+        <label htmlFor="name">Please type your name</label>
+        
+        <input
+        id="name" 
+        type="text"
+        value={userName}
+        onChange={onInputChange}/>
+        <button disabled={userName === ''} onClick={onNextChange} className="btn">Next</button>
```
Thereese, line 20: It´s possible to click the button even though there is no input, so I think this disabled-function doesn´t work. I´m not sure why though, maybe it´s something simple as removing the spaces?

Thereese (approved):
The code works well, no big errors or bugs in the evaluation tool. It looks really great on mobile and works well on desktop. A possible improvement is to make it a bit bigger on desktop, but that´s more of a personal opinion:) 
Great job!

PR #325, opened 2022-03-20
Week 10 project - Jessica Sandler

jessand77:
Link to my deployed project: https://jessicas-survey-project.netlify.app


code/src/components/Question2/Question2.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+
+import './Question2.css';
+
+const lostItems = ['suitcase', 'mind', 'dog', 'energy'];
```
nadialefebvre, line 5: Great that you created an array to map through instead of repeating the same piece of code below multiple times. Easier to add of remove items in it as well.

code/src/components/Question3/Question3.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import './Question3.css';
+
+const answers = ['today', 'yesterday', 'last week', "I don't know"];
```
nadialefebvre, line 5: Single or double quotation marks, we need to choose :) I just read today that it way more common in React to use single quotation marks, but in the jsx part (the HTML inside of these JS files, for exemple when you add the className attribute, or other attributes), it should be double quotation marks. You can set this so it's done automatically in VS code settings with Prettier. And then you just use the shortcut and your file has automatically the right ones (along with all the good formatting).

code/src/components/Question2/Question2.css
```diff
@@ -0,0 +1 @@
+/* Unnecessary file it seems */
```
nadialefebvre, line 1: Maybe you could remove it if it's not necessary?

code/src/components/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Footer = () => {
```
nadialefebvre, line 3: For some of your JS files (like this one), the indentation is 4 spaces instead of 2, so it makes the code unnecessarily wide (and we should follow the 2 spaces guideline).

nadialefebvre (commented):
Great job, Jessica! You can be proud of yourself! And I really appreciated your survey that confirmed that my mind is lost forever ;) I left some comments here and there, but it's really well done. Bravo!

PR #330, opened 2022-03-21
Joanna Lodell Week 10 Survey project

joannalodell19:
Link to my deployed project: https://the-alps-survey.netlify.app/

Due to sickness I had trouble with finishing the project this week. That's why I'll hand it in like this and then I'll come back to this and continue working on it. I think the app works as it should - besides my radio button. They can all be checked in, so I need to come back and fix that. Besides that my project needs styling.

### DIFF
--- a/code/src/components/Question2.js
+++ b/code/src/components/Question2.js
@@ -1,0 +1,37 @@
+import React from 'react'
+
+const Question2 = ({ goal,  onGoalChange, onStepChange}) => {
+    
+    const goals = [
+        'Getting some fresh air', 
+        'Working on my skiing skills', 
+        'Having a good time with friends', 
+        'Training for the Olympics']
+
+    return (
+      <div>
+            <article className="letter-container">
+                <form className="form">
+                    <h1 tabindex="0">What are you main goal for this ski trip?</h1>
+                    <div className="goal-input-container">
+                        {goals.map(goal => (
+                            <label className="radio-button-text" key={goal} tabindex="0">
+                                <input
+                                    className="radio-button"
+                                    type="radio"
+                                    value={goal}
+                                    onChange={onGoalChange}
+                                    tabindex="0"
+                                    >
+                                </input>
+                                {goal}
+                            </label>
+                        ))}
+                    </div>
+                   <button className='btn' type="button" onClick={onStepChange} tabindex="0">Next</button>  
+                </form>
+            </article>
+      </div>
+    )
+  }
+  export default Question2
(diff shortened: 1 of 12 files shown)


code/src/components/Question2.js
```diff
@@ -0,0 +1,37 @@
...
+                    <div className="goal-input-container">
+                        {goals.map(goal => (
+                            <label className="radio-button-text" key={goal} tabindex="0">
+                                <input
+                                    className="radio-button"
+                                    type="radio"
+                                    value={goal}
+                                    onChange={onGoalChange}
+                                    tabindex="0"
+                                    >
```
id4h4lling, line 25: I think you should add a checked to the index to fixed the button problem. 
Maybe:
checked={'type your Input' === 'type your value'} if i read it correct yours should be goal === goal. 
I hade different names for it.

code/src/components/Question3.js
```diff
@@ -0,0 +1,41 @@
...
+          <h1 tabindex="0">Choose your preferred weather</h1>
+          <div className="weather-input-container">
+            {weathers.map(weather => (
+                 <label className="weather-button-text" key={weather} tabindex="0">
+                    <input
+                      className="radio-button"
+                      type="radio"
+                      value={weather}
+                      onChange={onWeatherChange}
+                       >
```
id4h4lling, line 26: add the checked here as well.

code/src/components/Question3.js
```diff
@@ -0,0 +1,41 @@
...
+                    </input>
+                    {weather}
+                </label>
+                    ))}
+          </div>
+          
+          <button className='btn' type="button" onClick={onStepChange} tabindex="0">Next</button>
```
id4h4lling, line 33: Maybe instead of Next it should say submit, to indicate the last question. Or that is how I interpreted the Requirements.

id4h4lling (changes requested):
Good job on this project being sick and all! 👏
And I don't think you need to add too much more styling to be home. Some spacing and fonts and you're there. Sometimes less is more. (The radio buttons though that's a headache I dont wish upon a single soul, so if you do it and solve it please share 😉 ).

Over all nice and clean code that's easy to follow!
I added some comments on what i think is the solution for your radio buttons. If not give me holler and we can look at it together. 
And also a comment about the last button, which maybe should say submit according to blue level requirements. But since you used several pages maybe not really needed. 

Once again : nice work! 👏😊
/Ida

code/src/components/Question2.js
joannalodell19: Thanks, Ida - great help!

PR #331, opened 2022-03-21
Kristiina's survey project with react

kolkri:
https://application-survey-react.netlify.app/


code/src/Components/NameQuestion.js
```diff
@@ -0,0 +1,28 @@
...
+                    className="nextBtn button"
+                    disabled={nameInput===''}
+                    onClick={onStepChange}
+                    >Next <span className="arrow right"></span></button>
+            <div className="progress-container">
+                <div style={{width: `${progress}%`}} className="progress-value">
+                </div>
+            </div>
+        </div>
+      )
```
Mimmisen, line 27: You are very consistent and all of your code is written the same way, very clean!

code/src/Components/Form.js
```diff
@@ -0,0 +1,153 @@
...
+                            skillsArray={skillsArray} 
+                            linkInput={linkInput}
+                            experienceTitle={experienceTitle}
+                            onStepBackChange={onStepBackChange}
+                        />}   
+          {step === 9 && <ThanksPage
+                        />}  
+      </form>
+
+
```
Mimmisen, line 147: this is a very big file, you have a lot of questions so it would probably be nice to consider dividing the form as well as the components to make it cleaner/easier to find what you are looking for :)

code/src/Components/EmailQuestion.js
```diff
@@ -0,0 +1,33 @@
...
+            value={emailInput} 
+            placeholder="name@email.com"
+            onChange={onEmailInputChange}/>
+            
+            <div className="buttons">
+                <button className="button"
+                    disabled={emailInput===''}
+                    onClick={onStepChange}
+                    >Next <span className="arrow right"></span></button>
+                <button className="button"
```
Mimmisen, line 22: I like that you have added disabled on the buttons if the input is empty, but also adding disable to the email so that you only can add text with @ would be good

code/src/Components/Overview.js
```diff
@@ -0,0 +1,35 @@
...
+            <h3><span className="summaryTitle">Location:</span> {officeInput}</h3>
+            <h3><span className="summaryTitle">Skills:</span> {skillsArray.join(', ')}</h3>
+            <h3><span className="summaryTitle">Experience level:</span> {experienceTitle}</h3>
+            <h3><span className="summaryTitle">Link to portfolio:</span> {linkInput}</h3>
+        </div>
+        <div className="buttons">
+                <input 
+                    className="nextBtn button"
+                    type="submit"
+                    />
```
Mimmisen, line 22: I don't know how picky they are with this but it says in the blue level requirements: 
_"There should be a submit button. When pressed your app should hide the input form and show a summary of the user's submissions."_ 
but you have the submit in the summary which suits your survey better sins that would mean you send the application after you have looked at your answers first. but it's just an observation

code/src/Components/SkillsQuestion.js
```diff
@@ -0,0 +1,129 @@
...
+                    onClick={onStepBackChange}
+                    ><span className="arrow left"></span> Back</button>
+                
+            </div>
+            <div className="progress-container">
+                <div style={{width: `${progress}%`}} className="progress-value">
+                </div>
+            </div>
+        </div>
+      )
```
Mimmisen, line 128: this is also a very long component, if possible it would be better to find a way to make it shorter so you easier could make changes and more bug proof :)

code/src/Components/OfficeQuestion.js
```diff
@@ -0,0 +1,58 @@
...
+                <button className="button"
+                    onClick={onStepBackChange}
+                    ><span className="arrow left"></span> Back</button>
+                
+                </div>
+                <div className="progress-container">
+                <div style={{width: `${progress}%`}} className="progress-value">
+                </div>
+            </div>
+        </div>
```
Mimmisen, line 56: There is nothing wrong with this approach but if you want to you could watch the video from tech to learn with Jennie, she talks about creating an array with the different options and then mapping over them so that you don't need to make separate labels/inputs for each option and just use one for all of them! could be something to look up and try for the next time. :)

code/src/index.css
```diff
@@ -1,13 +1,416 @@
...
+  text-align: center;
+  margin:0;
+}
+h3{
+  margin: 0;
+  margin:0;
+}
+p{
+  margin: 0;
+  margin:0;
```
Mimmisen, line 63: here you have margin twice on both h3 and p

code/src/index.css
```diff
@@ -1,13 +1,416 @@
...
+    width: 1;
+  }
+}
+/* =====CONTAINER STYLES=====*/
+.container{
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  align-self: center;
```
Mimmisen, line 104: from what I can see/ and after trying it out in the inspector you don't need align-self here :)

code/src/index.css
```diff
@@ -1,13 +1,416 @@
...
+  opacity: 0.2;
+}
+
+.button:hover {
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.20), 0 6px 20px 0 rgba(0, 0, 0, 0.20);
+  color: #fff;
+	background-color: var(--red);
+	transform: scale(1.1);
+  cursor: pointer;
+}
```
Mimmisen, line 140: the indentions here is a bit uneven, would look  more clean if they were the same :)

Mimmisen (approved):
Great work this week on your project Kristiina! 🥇 
 You have made a beautiful survey that feels realistic. Your page is responsive, you can navigate with the keyboard and it was approved in lighthouse. So from what I can tell you have done the requirement! 💯  👍

PR #335, opened 2022-03-21
Project-Survey week 10 

Mimmisen:
Link to my deployed project:
https://code-student-survey.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,13 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <section className="app">
+      <div className="main-container">
```
tiiliu, line 8: From what I was able to notice, for the "app" you have a margin set in css and for the "main-container" you're setting the layout to display: flex and justify-content: center. I think you could combine this styling into one of these elements and then you could remove the other one and make the code a little bit cleaner that way :)

tiiliu (commented):
Good job Mimmi, you've created a really fun and nice survey! Love the funny memes 😁 
Your site works in all device sizes and is fulfilling the requirements concerning the inputs. There's some comments inside the code which I hope you will find helpful. One thing that I noticed was affecting all of your site was that when navigating the page with the keyboard, the hover styling wasn't working on many elements. So it wasn't possible to see whether the input or the button was chosen when moving with the tab key. The hover style worked again when using arrow keys to move inside the options and I was able to choose them and progress but you had to do it a little "blindly". So maybe that's something that would be good to look into for accessibility reasons? 😊  Perhaps using the focus attribute for this could work?
All in all, nice work! I enjoyed doing your survey and the pep talk in the end was lovely 🤗

code/src/components/FifthQuestion.js
```diff
@@ -0,0 +1,39 @@
...
+          Select one of the options
+        </label>
+        <select onChange={onGraduationChange} value={graduation}>
+          {graduate.map((item) => {
+            return (
+              <option key={item} value={item}>
+                {item}
+              </option>
+            );
+          })}
```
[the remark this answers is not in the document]
Mimmisen: Yes I was going for something a little fancier but I took your advice and changed the selector so now the "select one of the options" is disabled :)

code/src/App.js
Mimmisen: Thank you I made this change! :)

PR #340, opened 2022-03-24
Survey - Laura

Laura-Sjolander:
Here is the deployed survey: https://hybrid-worksurvey.netlify.app/


joannaringqvist (approved):
Great job with the survey! I think the code was clean and very well structured so it was easy to follow. Also thanks to the comments. Great that you had each question on its own page  and a previous/next button. Super! 🥳

PR #342, opened 2022-03-27
Dorothea's slaying survey

Dorothea-B:
Link to my deployed project:
https://slayer-survey.netlify.app/

I had a lot of fun with this project. I made things a bit hard for myself by being overly-ambitious with the CSS. It took way too much time to get everything to look right in both chrome, safari, firefox and edge, and there's still some bugs I didn't manage to fix. I tried for a lot of page transition animations as well, but at this point I just felt I had to publish without finishing.

Oftentimes when I'm googling for answers or browsing stack overflow, the solutions are using older syntax. It would be nice to get an explanation of react-classes, super(), "this.state" and stuff like that, so I can better understand and translate this into the newer syntax that we're using.


code/src/components/Form.js
```diff
@@ -0,0 +1,105 @@
...
+
+  const teamUpdate = (dude) => {
+    setShip(dude)
+}
+
+
+document.addEventListener('click', e => {
+  if (e.keyCode === '13' && step < 6) {
+    setStep(step +1);
+  } else if (e.keyCode === '37' && step > 1) {
```
kolkri, line 49: I was wondering what this else if statement is doing with keyCode 37, which seems to be the arrow left button, it should go one question back? It didn't work for me.

code/src/components/Form.js
```diff
@@ -0,0 +1,105 @@
...
+    setStep(step -1);
+  
+  }
+})
+
+  return (
+
+    
+      <div>
```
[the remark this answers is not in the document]
kolkri: I just noticed that you have used form tags inside of different components, which I think was on other way doing it, so never mind this comment :D

code/src/components/NameInput.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+export const NameInput = ({nameInput, nameUpdateInput, stepChange}) => {
+
+
+
```
kolkri, line 6: So you have used the form tags in the components, but you could also make the onSubmit  (instead of onClick) funtionality on them also.

code/src/components/Summary.js
```diff
@@ -0,0 +1,30 @@
...
+
+            <p>Name: {nameInput}</p>
+
+            <p> Shipping: {whatTeam === 'Angel' ? 'Team Angel' : 'Team Spike'}</p>
+
+
+            <p>Seen it: {watchedStatus}</p>
+            </div>
+
+            <div className='print-btn'><button onClick={window.print}>Print your results!</button></div>
```
kolkri, line 19: I cant see this button for some reason on the page

code/src/index.css
```diff
@@ -10,4 +10,198 @@ body {
...
+  position: absolute;
+  
+  width: 100vw;
+  height: 120vh;
+  padding: 0;
+  margin: 0;
+  background-color: black;
+  background: linear-gradient(-45deg, #781c00de, #ff0062c9, #115269cb, #000000f3);
+  background-size: 400% 400%;
+  animation: gradient 5s ease infinite;
```
kolkri, line 42: this animation is super cool!!

code/src/index.css
```diff
@@ -10,4 +10,198 @@ body {
...
+    border-right: 5px solid transparent;
+    top: .3em;
+    right: .75em;
+    border-top: 8px solid black;
+    opacity: 0.5;
+}
+select::-ms-expand {
+    display: none;
+} */
+
```
kolkri, line 201: you could remove the commented out code if you are not using them

code/src/index.css
```diff
@@ -10,4 +10,198 @@ body {
...
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  min-height: 500px;
+  background-image: url("https://i.ibb.co/q0sWp2q/pergament.png");
+  background-size: contain; 
+  background-repeat: no-repeat;
+  background-position: center;
+  z-index: 10;
+
```
kolkri, line 103: the pergament background are really nicely done!

code/src/styling/media-queries.css
```diff
@@ -0,0 +1,243 @@
+
+
+@media print {
+    body {
+  color: yellow;
+  background-color: greenyellow;
+    }
+  
+  }
+
```
kolkri, line 10: hmm i never seen this '@media print' syntax, so im not completely sure what this is doing

code/src/styling/range-slider.css
```diff
@@ -0,0 +1,130 @@
...
+    background: radial-gradient(circle, #db1616, #8b0000, rgb(97, 2, 2));
+  }
+  input[type=range]:focus::-ms-fill-upper {
+    background: radial-gradient(circle, #db1616, #8b0000, rgb(97, 2, 2));
+  }
+  
+  body {
+    padding: 30px;
+  }
+
```
kolkri, line 130: wow you have put a lot of effort for styling the range slider, and it is very cool! good thing that you have divided the css to different files, so it is easier to follow.

kolkri (commented):
Your page is really nice, all the details with design and animation makes it really cool! There are some parts of the code that I commented more. You have used forms for each question, but you never submit them, I think this point it is not even necessary since we dont collect the data, but later it is good to learn how to do the onsubmit thing as well.

code/src/components/Form.js
Dorothea-B: You're correct, I wanted a function for going backwards if pressing the left arrow key and forwards when pressing enter, but I didn't get it to work, should remove this.

code/src/components/Summary.js
Dorothea-B: My bad - wanted to make like a "profile card" for printing as a fun thing, using the @media print -query function, but didn't have time to finish it so i just disabled all of it.

code/src/styling/media-queries.css
Dorothea-B: It's styling not for a specific browser width but rather for how the page should look like when printing on paper. (So for a recipe for instance you might disable the navbar and ads and only print the recipe). But this i never finished.

PR #352, opened 2022-09-30
Linneas Neontastic Survey

LinneaWa:
Link to my deployed project: https://linneas-neontastic-survey.netlify.app/


code/src/components/Summary.js
```diff
@@ -0,0 +1,18 @@
...
+import arrow from './listarrow.png'
+
+// I wanted to make a bullet list with images instead of dots using list-style-image.
+// But list-style-image doesn't work with display:flex and I couldn't
+// align the text right without it so I ended up putting the image in the list.
+
+export const Summary = ({ name, home, lenght, bike, want }) => {
+  return (
+    <ul className="summary">
+      <li><img src={arrow} alt="arrow icon" />Your name is {name}.</li>
```
lindanorberg, line 11: Maybe one way to get the code slightly shorter would be to do a const for the image? Or maybe some kind of map-function? Not sure how though, sorry!

lindanorberg (commented):
So nice website! The neon and glowing effects are awesome! Great to do something fun when we don't have a design to follow :) 

I did understand most of the code easily, the one tiny part I didn't understand is becaues of my limited knowledge and not by poorly written code. Se comment below.  The survey seems to work like it should, the only thing is the scale input that doesn't go to the minimum or the maximum somehow. 

I understood that you were supposed to be able to read the page with a screen reade, but not sure if that was part of the requirement or not. The screen reader is not completely reading all the text. I am able to access the site with only the keyboard but the tabs are not always in the right order. Other than that the project meets the general requirements. 

Great work!!

PR #357, opened 2022-10-01
Project-survey - Tina Bruce

brucette:
Link to my deployed project:

https://grand-custard-84accf.netlify.app/


code/src/index.css
```diff
@@ -1,13 +1,218 @@
...
+  border: none;
+  border-radius: 4px;
+  color: #566b78;
+  font-size: 1rem;
+  font-weight: 600;
+  height: 3.9rem;
+  background: #e5e9ff;
+  width: 100%;
+  padding:  20px;
+  cursor: pointer;
```
malmen237, line 122: I find it really interesting to see the use of rem this extensively! I'm not comfortable with it so I don't use it at all, but it's been very educational to see it used. But maybe try not to mix the use of rem and px like this, although I don't know if there's some problems with rem in this case.

malmen237 (approved):
A very nice looking site! ⭐   
I find it easy and comfortable to navigate through, both in the browser and in the code.
Good job!

PR #359, opened 2022-10-01
Sarah's Survey

Kneedler:
Link to my deployed project: https://questions-about-food.netlify.app/

My reflections on how this week's project turned out: It was hard, Im not totally happy with it, but still - i managed to make it. I could not figure out how to put tabIndex on the thumb on the slider tho. 

Things I'd like to have clarified or explained in more detail: I haven't fully understood the eslint and how it works. And it puts {} in some places where I then put "", so the difference between them..


code/src/components/Summary.js
```diff
@@ -0,0 +1,106 @@
...
+  const onPreviousQuestionChange = () => {
+    setStep(step - 1);
+  };
+
+  const onEnter = (event) => {
+    if (event.key === 'Enter') {
+      event.preventDefault();
+      setStep(step + 1);
+    }
+  };
```
Scarylie, line 47: I dont think you need to have the if statement here for it to work.   This is how I did it and it works.

 ``` const onEnter = (event) => {  event.preventDefault();  setStep(step + 1); };```

code/src/components/PreferenceQuestion.js
```diff
@@ -0,0 +1,45 @@
...
+            <p>Hot food</p>
+          </label>
+          <label className="label-input-preference" htmlFor="preference">
+            <input
+              id={preference}
+              type="radio"
+              value="cold"
+              onChange={onPreferenceChange}
+              checked={preference === 'cold'}
+              onKeyPress={onEnter}
```
Scarylie, line 31: On your forms you could have called the onStepChange instead of making a new function called onEnter since they are doing the same thing. So it would look like this:
```onKeyPress={onStepChange}```

File: code/src/components/Summary.js [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+  }
+  const onHungerInputChange = (event) => {
+    setHungerInput(event.target.value)
+  }
+  const onStepChange = () => {
+    setStep(step + 1)
+  }
+  const onPreviousQuestionChange = () => {
+    setStep(step - 1);
+  };
```
Scarylie, line 40: I like this function. Might use it too now after seeing it here.

code/src/components/Overview.js
```diff
@@ -0,0 +1,28 @@
...
+
+const Overview = ({ nameInput, preference, sortOfPasta, otherIngredient, hungerInput }) => {
+  return (
+    <div className="survey-container">
+      <section className="overview-container">
+        <h2 tabIndex="0">Thanks for participating {nameInput}!</h2>
+        <div className="overview-summary">
+          <p tabIndex="0">You are in the mood for some {preference} food</p>
+          <p tabIndex="0">You prefere {sortOfPasta}, with some {otherIngredient}, nice!</p>
+          <p tabIndex="0">{hungerLevel[hungerInput]} <a href="https://www.yummly.com/recipes?q=pasta&taste-pref-appended=true" target="_blank" rel="noreferrer">Recipes!</a></p>
```
Scarylie, line 18: I like this idea that you give a link to recipies. Would be fun to get some different recepies depending on on the other input andon how hungry you said you were.

code/src/components/NameQuestion.js
```diff
@@ -0,0 +1,33 @@
...
+        <lable tabIndex="0" htmlFor="nameInput">
+          <input
+            id="nameInput"
+            type="text"
+            placeholder="Enter name here"
+            value={nameInput}
+            onChange={onNameInputChange}
+            onKeyPress={onEnter}
+            autoComplete="off"
+            className="name-input" />
```
Scarylie, line 23: To make input required just add ```required``` Like this:
```  <input  id="nameInput" type="text" placeholder="Enter name here" value={nameInput}  onChange={onNameInputChange} onKeyPress={onEnter}  autoComplete="off" className="name-input" required />```

Scarylie (approved):
Well done, your code is structured and your properties and functions are named good. I made some comments about how to make your code a little bit shorter and also how to add some requirements. Keep it up 🍝

code/src/components/Overview.js
Kneedler: I knooow! I wanted to do it a lot differently, I wanted to have if/else or something like that in every question.... but it got to komplex, I really tried figuring it out tho.

code/src/components/NameQuestion.js
Kneedler: Did that work for you? Because i had required in all of them, but it never worked, so i removed it..

code/src/components/PreferenceQuestion.js
Kneedler: Ooh, I have to try it out! Thank you for all the good feedback!

PR #360, opened 2022-10-01
Project Survey – Fredrik H

karlfredrikholm:
Link to my deployed project:
https://webdevsurvey.netlify.app/


Earnieball:
Not so much to add more than this was probably the most impressive work of the week. Really sick job, well done!

code/src/index.css
```diff
@@ -1,13 +1,44 @@
-body {
+*,
+*::before,
```
Earnieball, line 2: what do these pseudo-elements do?

karlfredrikholm: Since I'm not using any ::before and ::after in the rest of my code I don't THINK it makes any difference. I could have just written *. I've just seen people write it like that and I thought better safe than sorry. 

I have never used those pseudoelements myself yet, but they can for example be used for adding content before and/or after an element:
https://developer.mozilla.org/en-US/docs/Web/CSS/::before
https://developer.mozilla.org/en-US/docs/Web/CSS/::after

PR #361, opened 2022-10-01
Project Survey - Antonella Cardozo

Sailornina:
Link to my deployed project: https://survey-antonella.netlify.app/

### DIFF
--- a/code/src/components/SecondQuestion.js
+++ b/code/src/components/SecondQuestion.js
@@ -35,7 +35,7 @@ const SecondQuestion = ({ nextQuestion, onSecondQuestionChange }) => {
       <div className="content-wrap">
         <h1>Travel Sustainably</h1>
         <p>What is your most chosen transport?</p>
-        <p>Just choose one option! ⚠️</p>
+        <p>Just choose two options! ⚠️</p>
         <form className="form-checkbox">
           <label htmlFor="bus">
             <input
--- a/code/src/components/SecondQuestion.js
+++ b/code/src/components/SecondQuestion.js
@@ -35,6 +35,7 @@ const SecondQuestion = ({ nextQuestion, onSecondQuestionChange }) => {
       <div className="content-wrap">
         <h1>Travel Sustainably</h1>
         <p>What is your most chosen transport?</p>
+        <p>Just choose one option! ⚠️</p>
         <form className="form-checkbox">
           <label htmlFor="bus">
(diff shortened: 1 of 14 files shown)


code/src/components/SecondQuestion.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const handleWalkingInputChanged = (e) => {
+    handleTransportChoicesChanged(e, 'walking')
+  }
+
+  const handleOnClick = () => {
+    onSecondQuestionChange(transportChoices) // Fix me!
+    nextQuestion()
+  }
```
Fonnt, line 32: You could probably simplify it using .map() function

code/src/components/WelcomeToTheSurvey.js
```diff
@@ -0,0 +1,31 @@
...
+            <input
+              type="text"
+              onChange={onWelcomeChange}
+              value={nameWelcome}
+              id="name"
+              name="name"
+              placeholder="Your name" />
+          </label>
+        </form>
+        <button type="button" onClick={nextQuestion} className="button" disabled={nameWelcome === ''}>
```
Fonnt, line 23: Good job with the disabled. I've never used it. How does this work? Is this for screen readers?

code/src/components/WelcomeToTheSurvey.js
```diff
@@ -0,0 +1,31 @@
...
+        </p>
+        <form onSubmit={(e) => e.preventDefault()}>
+          <label htmlFor="name">
+            <input
+              type="text"
+              onChange={onWelcomeChange}
+              value={nameWelcome}
+              id="name"
+              name="name"
+              placeholder="Your name" />
```
Fonnt, line 20: You could probably increase accessibility a little more typing : "Type your name here"

Fonnt (commented):
Hi Antonella!
First of all I would like to say: Good job! Your project looks nice and I really liked going through your code! You accomplished all requirements and a little extra making it responsive. Nice!

I found interesting the way you approached the problem. I had tho small problems trying to understand the "flow". I needed to jump back and forward into all components to understand it. It would have been nice to have some comments explaining where the things came from. 

I think that keeping the App.js as simple as possible , is the way to go. Your Form component is also really clean and your code has a nice structure!. One can see that you spent time in that.

Keep doing! //David

PR #364, opened 2022-10-01
Project: Survey - Elin Segelöv

ElinSegelov:
Link to my deployed project:
https://benevolent-kitten-f30a19.netlify.app/

I'm pretty satisfied, more with the react code than with the design.

When and where to use ARIA attributes and tab indexes


code/src/components/RadioButton.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+const RadioButton = ({ valuedTrait, setValuedTrait, radioSelection }) => {
+  return (
+    <label className="radio-label" htmlFor={radioSelection}>
```
NaghmehOkhovat, line 5: Adding tabindex = "0" here after <label or in your ValuedTrait may help to make the tab work when using the Screen Reader. Right now I see that Tab key just work on the first button.

NaghmehOkhovat (commented):
Really good job with coding and also the design part specially with Harry Potter magic wand stick 💯 Your survey has everything that required as radio buttons, drop down window as well as the summary page. Just adding tabindex may help to have accessibility for the buttons too. I really like your idea for this project, survey and even all your previous projects. Good Luck! :)

PR #365, opened 2022-10-01
Project survey - Jessica Hansson

jesshansson:
Link to my deployed project:
https://find-a-friend-survey.netlify.app/


brucette (commented):
You have made a fun and professional looking survey! The hover effects on the buttons, box shadows and the ability to go back with the buttons all make it look finalised and polished - you can tell you had fun with it!  It functions as intended in all the different browsers I tested as well. 

I didn't really find anything to critique in the code so haven't entered any specific comments into it, it's very structured and clean, easy to follow - didn't find anything extra or unnecessary in it. 
The only thing I though about is that there is a similar function for event handling in all of your components, so I wonder if it would be possible to have just one general one in the App.js, e.g. called const handleEventChange(event), and then have that as a prop in all of the components instead.. though I didn't have time to test it so it may be simpler to have it as you have done, it is just a thought. 
Very nice work Jessica!

PR #366, opened 2022-10-01
Survey - Charlotte Johansson

charlottejohansson:
Link to my deployed project:
https://booksurvey.netlify.app/

This week was fun, I did not have that much time due to work but manage to hand in a project I'm happy of. I still find it a little hard to understand everything in React but like it more for every week :)

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -1,9 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
+import QuestionText from 'Components/QuestionText.js';
+import Header from './Components/Header.js';
+import QuestionDropdownGenre from './Components/QuestionDropdownGenre'
+import QuestionDropdownNumber from './Components/QuestionDropdownNumber'
 
 export const App = () => {
+  const [genre, setGenre] = useState('')
+  const [number, setNumber] = useState('')
+  const [recommendation, setRecommendation] = useState('')
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
+      <Header />
+      <section className="survey-container">
+        <QuestionDropdownGenre genre={genre} setGenre={setGenre} />
+        <QuestionDropdownNumber number={number} setNumber={setNumber} />
+        <QuestionText recommendation={recommendation} setRecommendation={setRecommendation} />
+      </section>
+    </>
   );
 }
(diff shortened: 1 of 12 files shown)


code/src/Components/QuestionRadio.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable */
+import React from 'react';
+
+const QuestionRadio = (props) => {
```
petrasson, line 4: Maybe there is a reason for doing it the way you do, but I think you can just add your props like this:
const QuestionRadio = ({ radio, setRadio } ) => {

code/src/App.js
```diff
@@ -1,9 +1,68 @@
...
+  const [showSummary, setShowSummary] = useState(false);
+  const [showForm, setShowForm] = useState(true);
+
+  const handleSubmit = ((event) => {
+    event.preventDefault()
+    setShowSummary(true)
+    setShowForm(false)
+  })
+
+  const surveyAnswered = () => {
```
petrasson, line 25: How does this work? Is it true if the question is an empty string? I would guess the opposite but I don't understand this true /false thing.

code/src/Components/Summary.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+function refreshPage() {
+  window.location.reload(false);
+}
+
+const Summary = (props) => {
```
petrasson, line 7: The same here (and on all props): You don't need to write props and deconstruct them into variables, just add them in the ({radio, genre, number, recommendation})

code/src/App.js
```diff
@@ -1,9 +1,68 @@
...
+      <Header />
+      <div>
+        {showForm &&
+          <form className="survey-container" onSubmit={handleSubmit}>
+            <QuestionRadio radio={radio} setRadio={setRadio} />
+            <QuestionDropdownGenre genre={genre} setGenre={setGenre} />
+            <QuestionDropdownNumber number={number} setNumber={setNumber} />
+            <QuestionText recommendation={recommendation} setRecommendation={setRecommendation} />
+            <div className="submitBtn-container">
+              <button className="submitBtn" disabled={surveyAnswered()} type="submit">Submit</button>
```
petrasson, line 52: Ah I think I get it. It's disabled as default (true) until it get false.

If that is the case, NICE ONE!

petrasson:
Great job Charlotte! I am very impressed by the design and colors. Looks really professional.
The survey works fine in different browsers and you have met all requirements.
Also clear naming and easy to follow the code. 

I added some comments mainly out of curiosity, but I think you can optimase some of the code if you like (see comment).
I think I will go back and have a look at this code whenever I feel lost in react. :)
WELL DONE!

PR #367, opened 2022-10-01
Neah Rockstroh, project-survey

Archkrull:
https://sport-survey-neah.netlify.app/

I definitely would have liked to have cleaner code and reusable components, but for my first project in React  and with  having less time this week than normal, I am happy with how it turned out.


code/src/App.js
```diff
@@ -1,9 +1,78 @@
...
+import { StartButton } from 'components/StartButton';
+import { Result } from 'components/Result';
+import { WeeklyTraining } from 'components/WeeklyTraining';
+import { PreviousButton } from 'components/PreviousButton';
+import { NextButton } from 'components/NextButton';
+import { SubmitButton } from 'components/SubmitButton';
+import { Intro } from 'components/Intro';
+import { AgeGroup } from './components/AgeGroup';
+import { YesNoQuestion } from './components/YesNoQuestion';
+import { SportType } from './components/SportType';
```
LinneaWa, line 13: It's really interesting to see how you have made components for the buttons. I did the buttons in a very basic way, without even one component, so thank you for showing me something different. I will probably try this next time. :)

code/src/components/AgeGroup.js
```diff
@@ -0,0 +1,34 @@
...
+                value={group}
+                onChange={handleAgeGroupChange}
+                checked={group === ageGroup} />
+              {group}
+            </label>
+          );
+        })}
+      </form>
+    </div>
+  )
```
LinneaWa, line 33: You say you wish your code was cleaner and not repetitive but here, for example, you have used .map in an excellent way to reduce number of lines.

code/src/index.css
```diff
@@ -1,13 +1,153 @@
...
+  transition: all 0.3s;
+  border-top-width: 1px;
+  border-bottom-width: 1px;
+  border-top-style: solid;
+  border-bottom-style: solid;
+  border-top-color: rgba(0, 0, 0, 0.5);
+  border-bottom-color: rgba(0, 0, 0, 0.5);
+  transform: scale(0.1, 1);
+}
+
```
LinneaWa, line 96: Really good job with the effect on buttons here!

LinneaWa (commented):
Great job! It's a nice looking survey and even though you had little time you still did some nice details to add the "cherry on top" for the user experience.  I understand you code and even if you think the button-component was a little repetitive I think you have a clean and readable code. The only thing I miss is some more comments in the code for some guidelines.

PR #368, opened 2022-10-01
Project Survey "Dining Up"

fannycardell:
Link to my deployed project:
https://dining-up-survey.netlify.app/


code/src/components/Email.js
```diff
@@ -0,0 +1,37 @@
...
+  <div className="sendEmail">
+    <input
+          className="emailInput"
+          type="text"
+          onChange={handleEmailAdressChange}
+          value={emailAdress} />
+          <button className="emailButton" type="button" onClick={handleConfirmEmailChange} tabIndex="0">Email Confirmation</button>
+    </div>
+
+  export default Email
```
Tanmark, line 37: This is awsome, and it is exported but not imported in the app.js so it doesn't come up as far as I can see in the survey

code/src/components/SuggestedResturant.js
```diff
@@ -0,0 +1,17 @@
...
+
+  return (
+
+    <><h2>We therefor suggest you to book a table for two at {suggestedResturant}</h2><input
+      type="text"
+      onChange={handleSuggestedResturantChange}
+      value={suggestedResturant} />
+    </>
+  );
+}
```
Tanmark, line 17: This one is the same as email, not imported to the app so I can't see it in the survey.

code/src/index.css
```diff
@@ -1,13 +1,206 @@
...
+.mainContainer {
+  display: flex;
+  flex-direction: column;
+}
+
+.surveyContainer {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
Tanmark, line 28: For the whole CSS, nice and neat code. Nicely done with the media queries, even if the styling might work better for a mobile screen, the text changes size so it is easy to read even on a larger screen. Well done!

Tanmark (commented):
I pointed out some of the components that did not work. Which of course means that all the other did work 🥳 I also saw in your readme that the reason for all the things I pointed out was time restraint so - I hear you! Your code is easy to read and nicely layed out. I think your survey is well built and works according to the brief. The styling is working best for mobile first which I think is the most important.

PR #372, opened 2022-10-02
Survey, Jessika Lind

jessikalind:
Link to my deployed project: [](https://stockholm-museum-survey.netlify.app/ )


code/src/Components/TextAreaInput.js
```diff
@@ -0,0 +1,30 @@
...
+const TextAreaInput = ({headline, input, setInput, inputLabel, id, placeholder, htmlFor}) => {
+
+  const handleInputChange = (event) => {
+    setInput(event.target.value)
+  }
+
+    return (
+      <>
+      <h2>{headline}</h2>
+        <form>
```
fannycardell, line 12: Is it maybe "event.preventDefault()" that is needed for the forms to reset?

code/src/Components/Result.js
```diff
@@ -0,0 +1,26 @@
...
+              <li><span className="bold-p">Favoriture museum:</span> {museum} </li>
+              <li><span className="bold-p">Museum memory:</span> {memory}</li>
+              <li><span className="bold-p">Possible improvemets:</span> {improvements}</li>
+            </ul>
+
+          <div className="link-container">
+            <span className="arrow" aria-hidden="true"> ✨ </span>
+            <a href="https://www.stockholmmuseums.se/en" target="_blank">Discover more about the Stockholm museum scene</a>
+            <span className="arrow" aria-hidden="true"> ✨ </span>
+          </div>
```
fannycardell, line 21: Nice idea to add a link to all the museums!

fannycardell (commented):
Hey Jessika! 
Great job with your interactive survey, works super well!
I liked theme overall=) and the "go back" function, the flowting bar below and the link to the museums in the end! Also 
the simple and yet playful background!

One thing I noticed was when I was going back to do it again, the answers was not reset (The age started at what I choosed last time, the moveble bar on question 2 started at the same amount of "museums visited", the textarea still had my 
written answers from the first time I did it (question 4 and 5). I think it is the "event.preventDefault()" that you have to add to the forms in the components=)

As you said on your demo, it would be clearer if the bar would start more to the left 
when the survey starts, so the answering person don't think they missed a question before. 
But it's small thing, I'm just impressed you even have it there=)

Congrats to a nicely made survey!

PR #375, opened 2022-10-02
Project Survey - Tolkien wishlist 

CeciliaMichele:
Link to my deployed project: https://astonishing-halva-8728d8.netlify.app/

My reflections on how this week's project turned out: I thought the example project was really helpful, but the CSS was quite hard I think.

Things I'd like to have clarified or explained in more detail: Would have loved to get more help with the CSS in the lectures, how to think about it in bigger terms


code/src/App.js
```diff
@@ -1,9 +1,43 @@
...
+      {step === 2 && (
+        <Language language={language} setLanguage={setLanguage} inputLabel="What language do you want?" />
+      )}
+      {step === 3 && (
+        <Version version={version} setVersion={setVersion} inputLabel="Which version would you like" />
+      )}
+      {step >= 4 && (
+        <Result book={book} language={language} version={version} />
+      )}
+      {step < 4 && (
```
jesshansson, line 33: It would be nice if you had the time to make another component (another page) that you get to after pressing submit. Just a simple confirmation that the email has been sent 👍

code/src/index.css
```diff
@@ -5,9 +5,169 @@ body {
...
+        font-size: 20px;
+      }
+      
+      .submit:hover:after {
+        opacity: 1;
+        right: 10px;
+      }
+    
+    
+      @media (min-width: 1024px) {
```
jesshansson, line 168: I hope you have the time to make it responsive! (now  it's a bit hard to see in mobile view) I don't think it was a must for this week though 👍  I understand that the map would be pretty hard to fit though 😄

code/src/App.js
```diff
@@ -1,9 +1,43 @@
...
+
+        <div className="next-step">
+          <p>Current step: {step}</p>
```
jesshansson, line 36: I actually think this could be removed ("current step")

jesshansson (approved):
First of all - the survey works - only that truly is a good job! Secondly - I love the theme :) And I liked the survey, it was a good idea to make a wishlist that you can send 👍 Feels like it could be a real thing!

I also like the layout, it matches all through. I think it works with the background as it is (not showing the whole image). You could maybe add some padding to your textboxes, it makes it a bit easier to read. 

Your survey also seem good looking at accessibility, when I ran it through Wave. There are only some small things to fix when I looked ran it through the validator (https://validator.w3.org/).

You've achieved all the goals for this week 👏

PR #376, opened 2022-10-02
Project-survey Maria Westling

marwebdesign:
Link to my deployed project:


code/src/Components/Drink.js
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <form>
+      <div className="drinkComponent">
+        <h2>What drink would you like as a reward for surviving another day of coding?</h2>
+
+        <select
+          onChange={handleDrinkChange}
+          className="selectDropdown"
+          aria-label="Select drink"
+          tabIndex="0"
```
jessikalind, line 17: Like we talked about on the Monday meeting, I am so far from an expert in the area of accessibility etc. But I don't think that you need to use a tab-index here, since the Select is a native HTML element which automatically is targeted when using tab (just like buttons, links and input elements for example). The only reason to use a tab index here, would be to change the order in which the site is "tab-able". But from what I can see, that should not be necessary 🥳

code/src/Components/Drink.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return (
+    <form>
+      <div className="drinkComponent">
+        <h2>What drink would you like as a reward for surviving another day of coding?</h2>
+
+        <select
+          onChange={handleDrinkChange}
+          className="selectDropdown"
+          aria-label="Select drink"
```
jessikalind, line 16: Nice! I haven't seen "aria-label" before. I guess this is just a way of "hiding" the label (so that it is not written out above the drop-down but read aloud by the SR?). I'm wondering (I do not have the answer to this) if the options also require aria-labels? But on the other hand they are read aloud by the SR as it is, so I'm guessing there is no need.

code/src/Components/Header.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const Header = () => {
+  return (
+    <div className="headerComponent">
+      <h1>SURVEY</h1>
+      <h2>Tell us how coding has been for you today</h2>
+      <span
+        className="computerIcon"
+        aria-label="Computer icon">&#128187;
+      </span>
```
jessikalind, line 11: Using the screen reader, I noticed that it read out: "Computer icon, group, computer". So it actually seems that the SR (at least the one I'm using, actually is reading out the content of the Emoji 😮

code/src/Components/Mood.js
```diff
@@ -0,0 +1,31 @@
...
+          <label key={moodType}>
+            <input
+              type="radio"
+              aria-label="Pick mood"
+              tabIndex="0"
+              onChange={handleMoodChange}
+              value={moodType}
+              checked={mood === moodType} />
+            {moodType}
+          </label>
```
jessikalind, line 26: Great solution building an array and using the map function here!

code/src/Components/Mood.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <form>
+      <div className="moodComponent">
+        <h2>What are your feelings after today?</h2>
+        {moodPreferences.map((moodType) => (
+          // eslint-disable-next-line jsx-a11y/label-has-associated-control
+          <label key={moodType}>
+            <input
+              type="radio"
+              aria-label="Pick mood"
```
jessikalind, line 20: Again, I am wondering a bit about the aria-label. Since you are using map here, it makes the aria-label the same for all radio buttons. This is fine, and makes sense in a way! But when you move to another radio button, you do not automatically get the content of that radio button read out loud. Perhaps setting the aria-label to "Pick mood: ${moodType}" would solve this? 😀

code/src/Components/Result.js
```diff
@@ -0,0 +1,19 @@
...
+export const Result = ({ name, mood, drink }) => {
+  return (
+    <div className="resultComponent">
+      <h2>Thank you for answering, this is how you responded:</h2>
+      <div className="results">
+        <p>Name: {name}</p>
+        <p>Current mood: {mood}</p>
+        <p>I prefer to drink: {drink}</p>
+      </div>
+      <h3>PS. You are doing great! You should be proud of yourself.</h3>
```
jessikalind, line 12: ❤️

jessikalind, line 12: The semantic structure has throughout the site been great! But I wonder if it would be better here to use a bold p instead of an h3, since the use of the h3 sort of implies that something will follow?

code/src/Components/Result.js
```diff
@@ -0,0 +1,19 @@
...
+      <span
+        className="heartIcon"
+        aria-label="Heart icon">&#129505;
+      </span>
```
jessikalind, line 16: Same as my comment regarding the computer icon.

code/src/index.css
```diff
@@ -5,9 +5,224 @@ body {
...
+  line-height: 120%;
+}
+
+.computerIcon {
+  font-size: 80px;
+  text-align: center;
+  padding: 30px 0px;
+}
+
+/* Name component  */
```
jessikalind, line 49: Nice structure in the CSS!

jessikalind (approved):
Great work on week 6, Maria! You have made such a nice-looking survey (with great styling on the input-fields and buttons), which really made me smile and feel a little better about the day-to-day coding-struggles. 

I think you have a great structure to your code, and a logical set-up of the components (and a very nice and clean app.js!). I myself am struggling a lot with the accessibility, and have learnt a lot from looking at your code (for example using "aria-label" instead of "label"). I however have some thoughts (more like: things that I am unsure of and would like to learn) on the labelling etc. See my comments for more details. 

Great job! 🥳

PR #378, opened 2022-10-02
Project survey - Younas

YoTesfam:
https://jade-kangaroo-7e67e7.netlify.app/


code/src/index.css
```diff
@@ -5,9 +5,159 @@ body {
...
+  font-size: 22px;
+}
+
+input[type="radio" i] {
+  margin: 10px;
+  font-size: x-large;
+}
+
+.startpage {
+display: grid;
```
ElinSegelov, line 116: How come you are using grid for the .startpage? I cant see that you are setting the grid anywhere. Since you are not doing that I think it would probably make more sense to use flex och block for display.

ElinSegelov (commented):
You have a nice and appropriate background color and even though there are many different colors I find the text on top nicely readable mostly with the text box background. Maybe giving it a slightly higher opacity would be good for the contrast.
You have some empty lines in you css and I would have cleaned up the index.html a little, removing the help text.
It looks really nice and I like your suggestions to improve the workplace. You code is well structured and easy to follow. Well done!

PR #380, opened 2022-10-02
Project Survey -Lisa 

LisaVasic:
Link to my deployed project:


code/src/App.js
```diff
@@ -1,9 +1,53 @@
...
+import { Rate } from 'Components/Rate';
+import { Result } from 'Components/Result';
+import { Header } from 'Components/Header';
 
 export const App = () => {
+  const [step, setStep] = useState(1);
+  const [name, setName] = useState('');
+  const [sandwich, setSandwich] = useState('');
+  const [pancakes, setPancakes] = useState('');
+  const [rate, setRate] = useState('');
```
joeohm, line 15: If you want the initial value to show up under the slider, you can set the value of rate in the state to 5. Then you also won't need the defaultValue property (which apparently should not be used in a controlled form according to ESLint).

code/src/Components/Drink.js
```diff
@@ -0,0 +1,51 @@
...
+        <p className="question3">What would you like to drink?</p>
+        <div>
+          <div className="inputContainer">
+            <input
+              className="inputRadio"
+              type="radio"
+              id="drinkOne"
+              value="Coffee"
+              onChange={handleDrinkChange}
+              checked={drink === 'Coffee'} /> Coffee
```
joeohm, line 20: Here you can use the `<label>` tag for the inputs. This will have the added benefit that you can then also click on the entire label text to select that option, instead of having to hit the small radio input circle. For example:
```
         <label htmlFor="blueberry">
            <input
              type="radio"
              id="blueberry"
              value="Blueberries"
              onChange={handlePancakesChange}
              checked={pancakes === 'Blueberries'} />
              Blueberry please!
          </label>
```

code/src/Components/Rate.js
```diff
@@ -0,0 +1,27 @@
...
+        <input
+          className="slider"
+          type="range"
+          id="rateInput"
+          name="rate"
+          min="1"
+          max="10"
+          step="1"
+          value={rate}
+          defaultValue="5"
```
joeohm, line 21: Using defaultValue in the controlled range form produces this warning:

> Warning: Rate contains an input of type range with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled input element and remove one of these props. More info: https://reactjs.org/link/controlled-components

defaultValue can be removed if setting the rate variable in the state to an initial value

joeohm (commented):
Lisa has made a beautiful breakfast survey that makes you long for blueberry pancakes!

The app is fully functional, no bugs, and meets the requirements for this week. The code is cleanly written and easy to follow.
Don't forget to update the readme about the project.

I like the color theme, which is also present in the hover effect on the buttons and throughout the app. 
Well done on the project overall, and on getting the controlled components to work with React!

//Joel Ö.

PR #383, opened 2022-10-02
Which character from Lord of the rings are you? - Agnes week-6

AgnesAhlman:
Link to my deployed project:
https://thelordoftherings-project6.netlify.app/


karlfredrikholm:
Very good job Agnes!! Code is easy to follow even without comments 😄 . It's so fun to see different approaches when React is so new to all of us. I like that you made the radio buttons their own component. And using separate css-files for each component 👍 Maybe you could have put the main font in the index.css so you wouldn't have to repeat it every time you were going to use it. But since I haven't tried this approach myself I'm not 100% sure how it works 😅 . The music is great too! 🧙‍♂️💍👏

Tara78 (commented):
Writing name before moving on to the starting game was excellent IDEA, and u did it so weel. 

Good JOB :)

PR #387, opened 2022-10-02
Project Survey by Linnéa Ajger

LinneaAjger:


code/src/App.js
```diff
@@ -1,9 +1,109 @@
...
+        {counter === 1 && (
+          <>
+            <Sleep
+              sleepHabit={userSleepHabit}
+              setSleepHabit={setUserSleepHabit} />
+
+            <div className="button-group">
+              <Button
+                btnText="Back"
+                btnClickFunction={() => handleCounterButtonClick(false)} />
```
marwebdesign, line 49: Nice idea to have a button that takes you back to the previous question. Easy solution that does a lot for accessibility!

code/src/components/Activity.js
```diff
@@ -0,0 +1,31 @@
...
+          <option label="0 min/week" value="0 min/week">0 min/week</option>
+          <option label="<30 min/week" value="<30 min/week">Less than 30 min/week</option>
+          <option label="30-60 min/week" value="30-60 min/week">30-60 min/week</option>
+          <option label="60-90 min/week" value="60-90 min/week">60-90 min/week</option>
+          <option label="90-150 min/week" value="90-150 min/week">90-150 min/week</option>
+          <option label="150-300 min/wee" value="150-300 min/week">150-300 min/week</option>
+          <option label=">300 min/week" value=">300 min/week">More than 300 min/week</option>
+        </select>
+      </div>
+      <p>Your answer: {activityTime}</p>
```
marwebdesign, line 27: Great to see your reply, both as a user and while coding - helps to see if the function works!

code/src/components/Food.js
```diff
@@ -0,0 +1,29 @@
...
+
+      <h3>How much veggies/fruit do you eat a day?</h3>
+      <div>
+        <select onChange={handleUserFoodHabitChange}>
+          <option selected="true" disabled value="Select an">Select an option:</option>
+          <option label="0gr" value="0gr">I do not eat any fruit och veggies </option>
+          <option label="0-200gr" value="0-200g">No more than 200g</option>
+          <option label="200-400g" value="200-400g">200-400g</option>
+          <option label="400-600g" value="400-600g">400-600g</option>
+          <option label=">600g" value=">600g">More than 600g</option>
```
marwebdesign, line 22: On screen reader - on the ones where you have written out "gr", it actually reads out "gram", while the ones where you only have "g" it will only read out "g". Good to know that the screen reader understands some shortenings!

code/src/components/Sleep.js
```diff
@@ -0,0 +1,56 @@
...
+    setSleepHabit(event.target.value)
+    console.log(setSleepHabit)
+  }
+  return (
+    <div className="content">
+      <SleepIcon />
+      <h2>Sleep habits</h2>
+
+      <h3>How many hours do you sleep per night?</h3>
+      <div className="radio-buttons">
```
marwebdesign, line 15: To shorten this code you could do an array with all the options, and then .map through them. However I do think your way of doing it is a very good way to get an understanding of how this type of code works.

code/src/components/Activity.js
```diff
@@ -0,0 +1,31 @@
...
+  const handleUserActivityChange = (event) => {
+    setActivityTime(event.target.value)
+  }
+  return (
+    <div className="content">
+      <ActivityIcon />
+      <h2>Physical activity</h2>
+
+      <h3>How many minutes/week do you spend on moderate/vigorous-intensity activities?</h3>
+      <a href="https://www.hsph.harvard.edu/obesity-prevention-source/moderate-and-vigorous-physical-activity/">What is moderate-vigorous intesity?</a>
```
marwebdesign, line 14: Great detail with some extra material (:

marwebdesign (approved):
This is such a nice looking survey! Great job! It really works as it should, and you should be really proud. 

I know you mentioned something about the radio buttons and how they reacted when clicked outside of the actual button, and when I tried your page a few days ago it seamed like it had some problems with collecting the correct response. However, now when I try it again everything seems to works fine, so I guess you managed to fix that. Good job (: 

When running it through the wave.webaim.org it had some issues with contrast, maybe the background color and font color are a bit too similar. Too bad though, I really liked the combination of colors! 

Sorry for this late code review, and again, Great job! See some extra comments in the code itself.

PR #388, opened 2022-10-02
Survey by Linda Malm

malmen237:
Watch the site live at:

https://about-your-day.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,101 @@
...
+    // If the user has inserted name and pressed enter it returns a value of
+    // undefined, if this happens the user will be ushered to the next question.
+    if (element !== undefined) {
+      element.preventDefault();
+    }
+    if (step < 7) {
+      setStep(step + 1)
+    }
+  }
+
```
CeciliaMichele, line 36: Really short and neat code in the App.js, really well done!

code/src/App.js
```diff
@@ -1,9 +1,101 @@
...
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="outer-wrapper">
+      <div className="inner-wrapper">
+        <div id="nextStep">
+          {step === 1 && (
+            <Start
+              handleStepIncrese={handleStepIncrese} />
```
CeciliaMichele, line 43: Just a typo that doesn't really change anything, but i believe it's spelled Increase

code/src/App.js
```diff
@@ -1,9 +1,101 @@
...
+              question4={question4}
+              handleStepIncrese={handleStepIncrese} />
+          )}
+        </div>
+        <div className="end-page">
+          {step === 7 && (
+            <p>Thank you for your time!</p>
+          )}
+        </div>
+        {/* A footer that is visable on every page. */}
```
CeciliaMichele, line 93: Loved the footer with the accessability hovering effect!

code/src/components/Question1.js
```diff
@@ -0,0 +1,33 @@
...
+    setChosen(true);
+  }
+
+  // Enables the user to use the enter-key to progress.
+  const checkKeyPress = (event) => {
+    const { keyCode } = event;
+    if (keyCode === 13) {
+      handleStepIncrese(event, chosen);
+    }
+  };
```
CeciliaMichele, line 17: Nice detail which makes it more accessible

code/src/components/Question2.js
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <>
+      <div className="question-box">
+        <p className="question">What are you doing today?</p>
+        <label htmlFor="workday">
+          <input
+            className="label-box"
+            name={question2}
+            type="checkbox"
+            id="workday"
```
CeciliaMichele, line 23: Loved the multiple options, I tried that too without success.

code/src/components/Question4.js
```diff
@@ -0,0 +1,47 @@
...
+
+            <option disabled value="">Select an option:</option>
+            <option value="working">Working</option>
+            <option value="studying">Studying</option>
+            <option value="away">Going away</option>
+            <option value="unicorn">Getting a unicorn</option>
+            <option value="bath">Going bathing</option>
+            <option value="restaurant">Going out to dinner</option>
+            <option value="workout">Getting a workout</option>
+            <option value="other">Other</option>
```
CeciliaMichele, line 34: Fun with all the options!

CeciliaMichele (commented):
A really nice, well functioning site that works well on desktop, I however feel like the font is a little bit on the tiny side on mobile. The buttons are a little bit hard to click on mobile, so I think a bigger size would make it perfect!
You have a really nice structure and the code is so neat and easy to follow with really good comments on what's happening. 
I remember reading in the common pitfalls that we should not use names as "question 2" etc, since we then need to rename them if we want to switch order. I think it makes it very clear though.
I also think it's nice with the short CSS, and that you gathered all of the id's in one chunk. 
Really good job this week, so impressed!

PR #392, opened 2022-10-02
Project survey - Linus

erilinus:
Link to my deployed project: 
https://splendorous-conkies-20a956.netlify.app/

I felt like I didn't have enough time to work on the styling. 

How to deploy a react website and small things like adding an image to a react build and how to build a radio button form.


code/src/index.css
```diff
@@ -11,3 +16,166 @@ code {
...
+}
+
+.button:hover {
+  background-color: rgb(110, 0, 108);
+  color: white;
+  border: none;
+}
+
+/* Utility styles */
+.top-margin {
```
a-well, line 146: Clever to put the margins in utility classes like that

code/src/index.css
```diff
@@ -1,3 +1,8 @@
+* {
```
a-well, line 1: Really neat and easy to follow CSS file!

code/src/App.js
```diff
@@ -1,9 +1,71 @@
...
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="outer-wrapper">
+      <div className="inner-wrapper">
+        {/* Where we gather the message */}
+        {counter === 0 && (
+          <div className="form-wrapper">
+            <Header />
+            <RecieverNameInput recieverName={recieverName} setRecieverName={setRecieverName} />
```
a-well, line 37: I noticed that the required attribute doesn't seem to be working. I didn’t have time to try this out so I’m not 100% sure it’ll do the trick, but I think that adding a <form> tag that wraps all the form elements should fix it. (Note that if you would want to try this out, any other <form> tags in the code would have to be removed since they can’t be nested 🙂)

a-well (approved):
Fun idea and very well executed 😀 The survey looks great, especially considering the fact that you didn't have a lot of time to style it. The code works well and apart from an accessibility thing (missing labels) that WAVE remarked on, the project fulfills all the blue level requirements 👍

PR #397, opened 2022-10-03
Maria's electricity survey

MariaWellander:
Link to my deployed project:


code/src/App.js
```diff
@@ -1,9 +1,43 @@
-import React from 'react';
+import React, { useState } from 'react';
+
+import { Age } from 'components/Age';
+import { Gender } from 'components/Gender';
+import { Electricity } from 'components/Electricity';
```
Archkrull, line 5: Everything in the right order, very clean!

code/src/App.js
```diff
@@ -1,9 +1,43 @@
...
+import { Result } from 'components/Result';
 
 export const App = () => {
+  const [step, setStep] = useState(1);
+  const [age, setAge] = useState('');
+  const [gender, setGender] = useState('');
+  const [electricity, setElectricity] = useState('');
+
+  const handleStepIncrease = () => {
```
Archkrull, line 14: Could be in it's own component, but as it is very short code it almost works better here as it would be more work creating a whole new component for it.

Archkrull (commented):
Great job with the components! 
To reach the blue level you only need to add some radio buttons and a dropdown menu, but as you have a very clean structure in your code and a great skeleton already, there shouldn't be any problems adding that. 

I've also tried your code with a screen reader, and it works fine with it so you hould be proud!


All in all, very structured and understandlable code and it works as it should. Great job!

PR #403, opened 2023-03-16
project-survey

theresBL:


code/src/components/Adjective.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+
+const hornAdjectives = [
+  'golden',
+  'sparkling',
+  'shiny',
+  'colorful'
+]
+
+export const Adjective = ({ horn, setHorn }) => {
```
JosefinRobertsson, line 10: This would be easier to connect to what it does with a name with 'horn' in it, when surveying your code, even though it makes sense.

code/src/components/Story.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+
+export const Story = ({ fairyName, color, horn, loveAct, name }) => {
+  return (
+    <>
+      <container className="story-page">
+        <p>
+          <span className="big-letter-span">O</span>nce upon a time, in a magical land far, far away, lived a beautiful,
```
JosefinRobertsson, line 8: There's a pseudo-element you could have tried here, p::first-letter, to style the O. In general the text part of this component is a bit messy to look at. An alternative approach could have been to remove the class names, use a couple of divs and  nth-of-type to target specific spans. Then you would have needed to count though and it would have been way less easy to come back and make a color change, so I do see the point of this too. It's easy to know what each class is referring to with your good classnames.

code/src/index.css
```diff
@@ -1,13 +1,166 @@
...
 
-code {
-  font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
-    monospace;
+.name-span {
+  color: blue;
 }
+
+.name-span2 {
+  color: deeppink;
```
JosefinRobertsson, line 133: This color is hard to read against the background.

code/src/components/Adjective.js
```diff
@@ -0,0 +1,31 @@
...
+  const handleHornAdjective = (event) => {
+    setHorn(event.target.value)
+  }
+  return (
+    <form className="horn-form" onSubmit={(event) => event.preventDefault()} required>
+      Choose a magical state:
+      {hornAdjectives.map((hornAdjective) => (
```
JosefinRobertsson, line 17: It would have been a little easier to make a choice if the labels were clickable too.

JosefinRobertsson (commented):
Lovely and coherent styling, everything works together to create an overall feeling. There could be a bit more contrast and/or larger font-size for easier reading, and the scrollbar would be nice to hide. Sweet little detail that the Next Question-button lights up, and the sound effects also adds to the experience and fairytale-feeling. Smart choice!
I find your code to be very neat and orderly and easy to follow along. Good choices on how to split the code up in components, and everything is doing its job without overcomplicating. It gives the impression that you were in control of this project. Good job!

PR #411, opened 2023-03-17
Project survey by Leo Thunell

khleoth:
Built upon the bussiness idea from the intro week


Yu-Miao222:
All files are neat and tidy, I like the way you keep code very dry! the layout of site looks great, back/next buttons, progress bar all look good and work well, maybe a slightly larger font on the page might be better :) WELL DONE!!

PR #418, opened 2023-03-18
Project survey - week 6

camcron:
https://mountain-lodge-questionnaire.netlify.app/


code/src/App.js
```diff
@@ -1,9 +1,88 @@
...
+  const [name, setName] = useState('');
+  const [date, setDate] = useState('');
+  const [transportation, setTransportation] = useState('');
+  const [equipment, setEquipment] = useState('');
+  const [addInfo, setAddInfo] = useState('');
+  const [textbox, setTextbox] = useState('');
+  const formRef = useRef(null);
+
+  const handleStepIncrease = () => {
+    if (formRef.current.checkValidity()) {
```
jonsjak, line 25: I'm not familiar with this validation control but it seems to be working for all but the select menu. Check that.
Also can't you just have a: `setStep(step +1)` instead because it applies when step is 0 as well?

code/src/components/AddInfo/Textbox/Textbox.js
```diff
@@ -0,0 +1,26 @@
...
+      <label htmlFor="textbox">
+        <input
+          type="text"
+          id="textbox"
+          className="textbox"
+          name="textbox-free"
+          onChange={handleTextboxChange}
+          value={textbox}
+          required
+          maxLength={100} />
```
jonsjak, line 22: Nice control of user input (one of many examples)

code/src/components/Summary/Summary.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  return (
+    <section className="summary-wrapper" aria-labelledby="summary-heading">
+      <h2 id="summary-heading">
+        Thank you for your time <em aria-label={`Added name: ${name}`}>{name}</em>!
+      </h2>
+      <p>
+        We have noted that you are going to the lodge by{' '}
+        <em aria-label={`Selected transportation method: ${transportation}`}>{transportation}</em>, that you are interested in renting{' '}
```
jonsjak, line 27: What is this tag?

jonsjak (approved):
Nice design! Good color scheme and I like the transparacy of the question-div.
Page is responsive and looks good on different devices. Also passes validations without remarks. Well done!

Suggestions: 
- Consider adding some kind of back or return button, especially on the last page where you get stuck as a user otherwise.
- See also a few inline-comments

Nicely structured code. Very complex, but well written. You put a lot of effort in to this I can tell.  Be proud!

PR #421, opened 2023-03-18
Project survey – state and controlled forms

SofiaGerdmar:
Note! All answers are treated anonymously, since they disappear into thin air when the reset button is clicked. :)


IrupePozoGraviz (approved):
I love it it such a beautiful design, well done!
everything works nicely and works good on mobile!
Keep up the good work you are awesome!!!

PR #423, opened 2023-03-19
Ninas week 6 project

NinaWald:


noraterbocs (commented):
I looked through your JS code, it looks quite organised, kudos for that!
If I had to mention something, I think from a UX point of view the input fields with the selected answer printed in them are not necessarily needed. Also for the CSS I would suggest to copy a ready made design because it's quite hard to match colors otherwise.
I think you solved this week's task well and I like it that you kept the project well structured.

PR #428, opened 2023-03-19
The Dream Survey by Frida N

fridanordenlow:
https://the-dream-survey.netlify.app/


code/src/App.css
```diff
@@ -0,0 +1,108 @@
...
+  margin: 15px 30px 15px 30px;
+  padding: 25px;
+  border-radius: 20px;
+  /* From https://css.glass */
+  background: rgba(255, 255, 255, 0.45);
+  border-radius: 16px;
+  box-shadow: 0 4px 30px rgba(0, 0, 0, 0.1);
+  backdrop-filter: blur(6.3px);
+  -webkit-backdrop-filter: blur(6.3px);
+  border: 1px solid rgba(255, 255, 255, 0.12);
```
fikamate, line 23: Im not sure how to make comments on several commits in a file but this is general about the styling.. FIrst of all i really love your bg-image!! It really talks with the theme of the survey and it gets very dreamy <3 The contrasts to the text elements could be a little bigger, its hard to read the text especially the footer content. You could easily fix this by make the backdrop-filter: blur into maybe 80px instead then its much more readable.

code/src/App.css
```diff
@@ -0,0 +1,108 @@
+header {
+  display: flex;
+}
+
+.header-title {
+  font-size: 29px;
+  margin: auto;
+  padding-top: 20px;
+  color: #ECCCCF;
+  opacity: 80%;
```
fikamate, line 10: Its kind of hard to se the title, since the color contrasts are very small, so consider maybe adding a background with less opacity or a backdropfilter to the h1 then it will bigger contrast to the background, or change the text-color to a darker color!

code/src/components/PreviousButton.js
fikamate, line 1: I  really like that you made this work! Its very convienient as a user to be able to go back 🙌🏻

code/src/components/buttons.css
fikamate, line 1: Great job making it responsive!!

fikamate (approved):
Very nice work with this! I live the theme and how its all consistent! Everything works according to the instructions and the code is easy to follow. It could be added some more comments in the code but im fully aware of the problems it causes with eslint. The only changes suggestions i had was about the styling and contrasts (see comments!). Great job! :)

PR #433, opened 2023-03-19
Project survey

LinaAdamsson:


code/src/components/End.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+export const End = ({ name, yesno, shape }) => {
+  // console.log(name,yesno,shape, "--------")
+  if (yesno === 'No') {
+    yesno = 'Just like you';
+  } else if (yesno === 'Yes') {
+    yesno = 'Unlike you';
+  }
```
AndreaHedstrom, line 9: Smart to add an if/else if to change the text depending on the answers from the user!

code/src/components/Name.js
```diff
@@ -0,0 +1,14 @@
...
+    setName(event.target.value);
+  }
+  return (
+    <div className="name-container">
+      <div className="question">what is your name?
+        <input type="text" className="name-input" value={name} onChange={handleNameChange} />
+      </div>
+    </div>
+  );
+}
```
AndreaHedstrom, line 14: All code feels clean and easy to follow, good classnames etc

code/src/styles.css
```diff
@@ -0,0 +1,209 @@
...
+  font-size: 40px;
+  margin-top: 172px;
+  margin-left: -692px;
+  padding-left: 4px;
+  padding-top: -10px;
+  padding-right: 4px;
+  padding-bottom: 0px;
+  transform: rotate(-80deg);
+  border: 0.5px solid rgb(178, 175, 175);
+  }
```
AndreaHedstrom, line 31: Impressed that you got the buttons to be in their hands! It was a nice detail.

code/src/styles.css
```diff
@@ -0,0 +1,209 @@
...
+overflow: hidden;
+}
+
+.app-container span {
+  color: rgb(88, 80, 80);
+  background-color: white;
+  font-family: Manga;
+  align-items: center;
+  text-align: center;
+  font-size: 40px;
```
AndreaHedstrom, line 22: I did exactly this my self so well - for both of us - maybe try to write margins and paddings in one line when we use more than one of them (Like margin: 10px 5px 5px 10px;) It's just a bit hard to remember wich one do margins where I think.

AndreaHedstrom (commented):
Great job Lina! Like I said it is easy to understand and good classnames etc! 
I realised that the background-image didn't work on Safari for some reason - maybe that's something you want to fix when there is time :) 
Heja dig, bra kämpat!! <3

PR #440, opened 2023-03-19
My Little Pony name generator

fikamate:


code/src/App.js
```diff
@@ -1,9 +1,91 @@
...
+      {step === 5 && (
+        <MagicCrystalSelect
+        label="Select your favourite magic crystal"
+        Value={magicCrystalSelect}
+        setMagicCrystalSelect={setMagicCrystalSelect} />
+      )}
+      {step === 6 && (
+        <ToppingSelect
+        label="Select your favorite topping"
+        ToppingSelect={toppingSelect}
```
fridanordenlow, line 66: I think the uppercase in the ToppingSelect here is making the props not being passed down correctly? toppingSelect={toppingSelect}

code/src/App.js
```diff
@@ -1,9 +1,91 @@
...
+        setToppingSelect={setToppingSelect} />
+       )}
+      {step === 7 && (
+        <Submit Submit={submit} setSubmit={setSubmit} />
```
fridanordenlow, line 70: I can't see that you have any props in your Submit-component so you don't need to pass down any props here, you can just mount it like this: `<Submit />`

code/src/App.js
```diff
@@ -1,9 +1,91 @@
...
+import { ToppingSelect } from 'components/ToppingSelect';
+import Header from './components/Header'
+import Start from './components/Start';
 
 export const App = () => {
+  const [step, setStep] = useState(1);
+  const [name, setName] = useState('');
+  const [colorSelect, setColorSelect] = useState('');
+  const [magicCrystalSelect, setMagicCrystalSelect] = useState('');
+  const [submit, setSubmit] = useState('');
```
fridanordenlow, line 20: Adding to the Submit component comment further below, I don't think you need the submit and setSubmit

code/src/App.js
```diff
@@ -1,9 +1,91 @@
...
+        ToppingSelect={toppingSelect}
+        setToppingSelect={setToppingSelect} />
+       )}
+      {step === 7 && (
+        <Submit Submit={submit} setSubmit={setSubmit} />
+        )}
+      {step === 8 && (
+        // eslint-disable-next-line max-len
+        <Result name={name} colorSelect={colorSelect} magicCrystalSelect={magicCrystalSelect} fruitSelect={fruitSelect} toppingSelect={toppingSelect} />
+      )}
```
fridanordenlow, line 75: I don't seem to get the correct result showing after I have submitted my answers, some of theprops are passed down but not all of them. For example after Cutie mark: and Your secret pony weapon: it's blank. I'm not sure why, maybe it's the toppingSelect messing with things? I haven't found anything else that it could be

fridanordenlow (commented):
Hi Annika! A lovely choice of theme for your survey, very nicely done! You have quite many steps and most things are working  but there is something with the props being passed between components and the App that's not working? I couldn't get a complete result, see comments below. I think it worked when you demoed in our team session so I wonder what it could be? It's probably just something small causing the bug.

PR #444, opened 2023-03-19
Project Survey Josefin R

JosefinRobertsson:


code/public/index.html
```diff
@@ -13,7 +13,9 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Technigo React App</title>
+    <link href="https://fonts.googleapis.com/css2?family=Permanent+Marker&display=swap" rel="stylesheet">
```
MalinSkill, line 16: I think the two fonts you picked looks really good together.

code/src/App.js
```diff
@@ -1,9 +1,93 @@
...
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
+      <Header />
+      {progress < 1 && (
+        <div className="introMsg">
+          <h3>
+      You&apos;ve reached a survey made by your local convenient store.
```
MalinSkill, line 41: Nice to use this way to get &apos;

code/src/components/TasteGroup.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import SnacksArray from './SnacksArray';
+
+const TasteGroup = ({ tasteGroup, setTasteGroup }) => {
+  const uniqueTastes = [...new Set(SnacksArray.map((snack) => snack.taste))];
+
```
MalinSkill, line 14: This is really impressive, using one array to store all the different  snacks. And then make a new array of one specific value from that array. which is then mapped again for the buttons. Just typing it out to try and wrap my head around it.

code/src/App.js
```diff
@@ -1,9 +1,93 @@
...
+      We&apos;re about to stock our shelves with some new snacks and we&apos;d love to
+      get your opinion on what to put on them. Thank you for your time!
+            <p>-Here for you day & night, LocalMart</p>
+          </h3>
+        </div>
+      )}
+      {progress === 1 && <TasteGroup tasteGroup={tasteGroup} setTasteGroup={setTasteGroup} />}
+      {progress === 2 && <TypeOfSnack tasteGroup={tasteGroup} type={type} setType={setType} />}
```
MalinSkill: Great work, passing the values from earlier questions to be able to do the conditionals!

code/src/components/TypeOfSnack.js
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
+import SnacksArray from './SnacksArray';
+
+const TypeOfSnack = ({ tasteGroup, type, setType }) => {
+  const handleTypeChoice = (event) => {
+    setType(event.target.value);
+  };
+  const filteredForTasteGroup = SnacksArray.filter((snack) => snack.taste.includes(tasteGroup));
```
MalinSkill, line 23: So impressed by this! Filter to only include the values corresponding with previous answer and then making a new array based on that. Mind blowing!

code/src/App.js
```diff
@@ -1,9 +1,93 @@
...
+        setComment={setComment}
+        textInput={textInput}
+        setTextInput={setTextInput} />}
+      {progress <= 6 && <ProgressButton
+        progress={progress}
+        setProgress={setProgress}
+        tasteGroup={tasteGroup}
+        type={type}
+        texture={texture}
+        flavor={flavor}
```
MalinSkill: good idea to have the progress button be a separate component. But maybe it does not need all that is being passed to it?

code/src/index.css
```diff
@@ -1,13 +1,111 @@
...
+}
+
+
+label:hover {
+  transform: scale(1.1);
+  animation: shake 0.5s;
+  animation-iteration-count: infinite;
+}
+
+@keyframes shake {
```
MalinSkill, line 69: this is so cool for the buttons

code/src/components/Results.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const Results = ({ tasteGroup, type, texture, flavor, priceValue, textInput }) => {
+  const flavorsArray = flavor.length > 2
+    ? `${flavor.slice(0, -1).join(', ')} and ${flavor.slice(-1)}`
+    : flavor.join(' and ');
```
MalinSkill, line 6: Wow, I feel like you have really thought of everything! this looks like a great way to decide how it should be displayed depending on how many words there are

MalinSkill (commented):
Wow what an amazing project! the amount of different objects you have in your array is really impressive and how you make it work with filtering and mapping. The layout is responsive, I tried also on my phone. One thing to look into is that proceed button is very small on phone. Also when I went for salty soft candy I got licorice and fruit as flavour to choose from. Such an impressive and complex survey! great work!

PR #448, opened 2023-03-19
Oscar's Project Survey

OscarSindemark:
https://wondrous-gelato-35a604.netlify.app

In progress


code/src/App.js
```diff
@@ -1,9 +1,48 @@
...
+      {step === 4 && (
+        <Show show={show} setShow={setShow} />
+      )}
+      {step >= 5 && (
+        <Results name={name} genre={genre} movie={movie} show={show} />
+      )}
+
+      {step < 5 && (
+        <>
+          <p>Current page: {step} </p>
```
AnnikaSonnek, line 42: It was a nice feature that you the user could see which step they were on.

code/src/components/Assets/aot.jpg
AnnikaSonnek, line 1: I LOVE ATTACK ON TITAN! I was so happy to see it among the options! Hahah!

code/src/components/Movie.js
```diff
@@ -0,0 +1,40 @@
...
+          <img src={require('./Assets/arrival.jpg')} alt="Arrival Movie" />
+          <input type="radio" value="Arrival" />
+        Arrival
+        </div>
+        <div className="loveActually">
+          <img src={require('./Assets/love-actually.jpg')} alt="Love Actually Movie" />
+          <input type="radio" value="Love Actually" />
+        Love Actually
+        </div>
+        <div className="eeaao">
```
AnnikaSonnek, line 27: Love the name of this div! Hahah! Such a long movie title!

code/src/index.css
```diff
@@ -5,9 +5,23 @@ body {
     sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background-image: url('./components/Assets/movies-background.jpg');
```
AnnikaSonnek, line 8: I liked that you had an image as a background!

code/src/components/Show.js
```diff
@@ -0,0 +1,40 @@
...
+          <input type="radio" value="New Girl" />
+        New Girl
+        </div>
+        <div className="madMen">
+          <img src={require('./Assets/mad-men.jpg')} alt="The show Mad Men" />
+          <input type="radio" value="Mad Men" />
+        Mad Men
+        </div>
+        <div className="attackOnTitan">
+          <img src={require('./Assets/aot.jpg')} alt="The show Attack on Titan" />
```
AnnikaSonnek, line 33: What does "require" mean in this context? I've never seen it before!

code/src/App.js
```diff
@@ -1,9 +1,48 @@
...
+  const handleStepIncrease = () => {
+    setStep(step + 1)
+  }
+
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  );
+    <form className="surveyForm">
```
AnnikaSonnek, line 23: I see that you made a form instead of a div here, how come? 
Also, if you want to style it so that it's not as wide you can specify that with the max-width or width in the CSS.

AnnikaSonnek (commented):
I ran your website through the W3 Validator and there were no issues there. Yay! 

I also ran it through the WAVE accessibility tool and there where some contrast issues with the green text of the step counter, no issues with the header though. I think its because the header is bold. There were also some form elements missing around the select and text-input. I saw that you made the big container surveyForm into a form but maybe it would be good to have that one as a div instead and then add the form element in each of the components instead. 

I also checked the responsiveness and it's not responsive for mobile view. 

It would be nice with some comments maybe but the code is easy to follow without them as well. Good job!

PR #458, opened 2023-03-22
Project Survey - Andrea Hedström

AndreaHedstrom:


code/src/App.js
```diff
@@ -1,9 +1,51 @@
...
+          <div className="step">
+            <p>PARTYSTEP {step}/5</p>
+          </div>
+
+        </>
+      )}
     </div>
   );
-}
+};
```
beemailley, line 51: I like that your App.js is very tidy and organized.  It's very easy to see how everything is related.

code/src/index.css
```diff
@@ -11,3 +11,171 @@ code {
...
+  align-self: center;
+  align-items: center;
+  justify-content: center;
+  background-color:rgba(239, 154, 181, 0.458);
+  height: 600px;
+  margin-top: 50px;
+  margin-left: 25px;
+  margin-right: 25px;
+  border-radius: 10px;
+
```
beemailley, line 36: So, I won't comment a ton on your CSS - you've already said you're updating it.  But, I wanted to share "vw" and "vh" as measurements for view width and view height.  If you've already experimented with these, ignore me.  They were new to me as I was styling this project and were SUPER helpful in forcing my survey to take up the amount of space that I wanted it to.

code/src/components/Drink.js
```diff
@@ -0,0 +1,30 @@
...
+            And drink?
+        </h2>
+      </div>
+      <div className="select-container">
+        <select
+          className="select"
+          id="drink"
+          value={drink}
+          onChange={handleDrinkChange}>
+          <option value="">Menu</option>
```
beemailley, line 20: I think you can add "disable" either before or after "value" for the MENU line and that will stop being an acceptable value. Then the user would have to pick something else from the dropdown.

code/src/components/Food.js
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+      <div className="foodBtn">
+        {foodOption.map((group) => (
+          <label key={group} htmlFor="group">
+            <input
+              type="radio"
+              className="radioBtn"
+              value={group}
+              onChange={handleFoodChange}
+              checked={food === group} />
```
beemailley, line 27: I think an easy way to add a bit of accessibility is to add an aria-label here.  Something like:
`aria-label={group}`
Then the label should change for each radio button.

code/src/components/Party.js
```diff
@@ -0,0 +1,36 @@
...
+            (jsConfetti.addConfetti({
+              confettiRadius: 8,
+              confettiNumber: 500,
+              confettiColors: [
+                '#d44e62', '#fcfee1', '#eebed9', '#b4c8b7', '#fb91c6', '#a0bbd0'
+              ]
+            }))
+          }}>MAGIC PARTYBUTTON
+        </button>
+
```
beemailley, line 32: The party button is such a cute addition! And I like that the confetti colors match your design scheme.

beemailley (commented):
Overall, everything looks good! Your code looks very clean and very easy to read.  The components are easy to understand.  Good job!

